PR #5, opened 2021-01-26
Th2 1125

Nikita-Smirnov-Exactpro:


Dockerfile
```diff
@@ -13,7 +13,7 @@ RUN gradle --stacktrace clean build dockerPrepare artifactoryPublish \
      -Partifactory_deploy_repo_key=${artifactory_deploy_repo_key} \
      -Partifactory_url=${artifactory_url}
 
-FROM openjdk:12-alpine
+FROM adoptopenjdk/openjdk11:alpine
 ENV RABBITMQ_HOST=rabbitmq \
```
OptimumOpium, line 17: I think we can remove env variables from Dockerfile because we don't need them anymore

build.gradle
```diff
@@ -113,24 +113,17 @@ bintray {
...
     implementation "org.slf4j:slf4j-log4j12"
     implementation "org.slf4j:slf4j-api"
 
-    implementation "com.google.protobuf:protobuf-java-util"
-    implementation "io.grpc:grpc-protobuf"
-    implementation "io.grpc:grpc-stub"
-    implementation "io.grpc:grpc-core"
-    implementation "io.grpc:grpc-netty"
-
     implementation "com.exactpro.sf:sailfish-core:${sailfishVersion}"
```
OptimumOpium, line 124: Please, check the Sailfish version and replace it with a snapshot if it is a release version. Other boxes use snapshot version so I think we should use it here as well

OptimumOpium (approved):

OptimumOpium (approved):

PR #16, opened 2021-04-30
Th2 1863

Nikita-Smirnov-Exactpro:


src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -292,27 +316,84 @@ abstract class AbstractCheckTask(
...
+        var builder = factory.invoke()
+
+        eventGroups[eventID]?.forEach { event ->
+            val checkedEvent = checkAndRebuild(event)
+
+            LOGGER.debug("Process ${checkedEvent.name} ${checkedEvent.type}")
+            if(eventGroups.containsKey(checkedEvent.id)) {
+                builder.addEvents(checkedEvent)
+                result.add(checkAndBuild(builder))
+                builder = factory.invoke()
```
OptimumOpium, line 360: Is it possible if we have several events with the same parent ID and one of them has child events and another one doesn't?
If it is possible then we can have the following situation:
1) We process the event without child events -> add it to the builder (line 370)
2) We process the event with child events -> add it to the builder (line 358) -> convert builder to the batch (batch has 2 events)

In that case, we have the event that has child events in the batch with several events. And it is not allowed If I remember correctly. Please, correct me if I am wrong

OptimumOpium (changes requested):

src/test/kotlin/com/exactpro/th2/check1/rule/sequence/TestSequenceCheckTask.kt
```diff
@@ -111,10 +112,32 @@ class TestSequenceCheckTask : AbstractCheckTaskTest() {
...
+            with(batchRequest[2]) {
+                assertEquals(3, eventsCount)
+                assertTrue (getEventsList().all { "Verification" == it.type })
+            }
+            with(batchRequest[3]) {
+                assertEquals(2, eventsCount)
+                assertNotNull (getEventsList().firstOrNull { "checkSequence" == it.type })
+                assertNotNull (getEventsList().firstOrNull { "checkMessages" == it.type })
+            }
+            with(batchRequest[4]) {
```
OptimumOpium, line 136: I think we need to check the `batchRequest` size before getting the events by index

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
Nikita-Smirnov-Exactpro: Yes, you right. We should guarantee that event with children are published as a single event

src/test/kotlin/com/exactpro/th2/check1/rule/sequence/TestSequenceCheckTask.kt
Nikita-Smirnov-Exactpro: the awaitEventBatchRequest(timeout, times) method already check it

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -73,7 +68,7 @@ abstract class AbstractCheckTask(
 
     init {
         if (maxEventBatchContentSize <= 0) {
-            throw IllegalArgumentException("'maxEventBatchContentSize' can't be lower zero, actual: $maxEventBatchContentSize")
+            throw IllegalArgumentException("'maxEventBatchContentSize' should be greater than zero, actual: $maxEventBatchContentSize")
         }
```
OptimumOpium:
```kotlin
require(maxEventBatchContentSize > 0) {
    'maxEventBatchContentSize' should be greater than zero, actual: $maxEventBatchContentSize"
}
```

src/main/kotlin/com/exactpro/th2/check1/rule/CheckTaskUtils.kt
```diff
@@ -0,0 +1,101 @@
...
+import com.google.protobuf.TextFormat
+import java.lang.IllegalArgumentException
+import com.exactpro.th2.common.grpc.Event as ProtoEvent
+
+val OBJECT_MAPPER = ObjectMapper()
+
+fun Event.disperseToBatches(maxEventBatchContentSize: Int, parentEventId: EventID): List<EventBatch> {
+    if(maxEventBatchContentSize <= 0) {
+        throw IllegalArgumentException("'maxEventBatchContentSize' should be greater than zero, actual: $maxEventBatchContentSize")
+    }
```
OptimumOpium, line 34:
```suggestion
    require(maxEventBatchContentSize > 0) {
        "'maxEventBatchContentSize' should be greater than zero, actual: $maxEventBatchContentSize"
    }
```

src/main/kotlin/com/exactpro/th2/check1/rule/CheckTaskUtils.kt
```diff
@@ -0,0 +1,101 @@
...
+
+    val events = toListProto(parentEventId)
+
+    val result = ArrayList<EventBatch>()
+    val eventGroups = events.groupBy(ProtoEvent::getParentId)
+    batch(maxEventBatchContentSize, result, eventGroups, parentEventId)
+    return result
+}
+
+private fun batch(maxEventBatchContentSize: Int, result: MutableList<EventBatch>, eventGroups: Map<EventID, List<com.exactpro.th2.common.grpc.Event>>, eventID: EventID) {
```
OptimumOpium, line 44: You can add an alias for `com.exactpro.th2.common.grpc.Event` in import to avoid writing the fully qualified name

src/main/kotlin/com/exactpro/th2/check1/rule/CheckTaskUtils.kt
```diff
@@ -0,0 +1,101 @@
...
+    val factory = { EventBatch.newBuilder().setParentEventId(eventID) }
+
+    var builder = factory.invoke()
+
+    checkNotNull(eventGroups[eventID]) {
+        "Neither of events refers to ${TextFormat.shortDebugString(eventID)}"
+    }.forEach { event ->
+        val checkedEvent = checkAndRebuild(maxEventBatchContentSize, event)
+
+        AbstractSessionObserver.LOGGER.trace("Process ${checkedEvent.name} ${checkedEvent.type}")
```
OptimumOpium, line 54: We should not use the logger from different class

src/test/kotlin/com/exactpro/th2/check1/rule/sequence/TestSequenceCheckTask.kt
```diff
@@ -112,9 +112,17 @@ class TestSequenceCheckTask : AbstractCheckTaskTest() {
 
         sequenceCheckRuleTask(parentEventID, messageStream, checkOrder).begin()
 
-        val batchRequest = awaitEventBatchRequest(1000L, 5)
+        val batchRequest = awaitEventBatchRequest(1000L, 6)
```
OptimumOpium, line 115: Why did we change the number of invocations?

Nikita-Smirnov-Exactpro: The previous version has a bug related to union events with children and without children.
Please have a look schema
```
        /*
            checkSequenceRule
              preFiltering
                Verification x 3
              checkMessages
                Verification x 3
              checkSequence
         */
```
The previous version unions checkMessages and checkSequence events

README.md
```diff
@@ -79,6 +81,9 @@ The value will be interpreted as time unit defined in _cleanup-time-unit_ settin
 #### cleanup-time-unit
 The time unit for _cleanup-older-than_ setting. Available values are MILLIS, SECONDS, MINUTES, HOURS. _The default value is SECONDS_
```
jean-michel-perli-sonsin, line 82:
```suggestion
The time unit for _cleanup-older-than_ setting. The available values are MILLIS, SECONDS, MINUTES, HOURS. _The default value is set to SECONDS_
```

README.md
```diff
@@ -79,6 +81,9 @@ The value will be interpreted as time unit defined in _cleanup-time-unit_ settin
...
 
+#### max-event-batch-content-size
+The max size in bytes of summary events content in a batch defined in _max-event-batch-content-size_ setting. _The default value is 1048576_
+
 ## Requried pins
```
jean-michel-perli-sonsin, line 87:
```suggestion
## Required pins
```

README.md
```diff
@@ -79,6 +81,9 @@ The value will be interpreted as time unit defined in _cleanup-time-unit_ settin
...
 
 The Check1 component has got tow types of pin
```
jean-michel-perli-sonsin, line 89:
```suggestion
The Check1 component will have two types of pin:
```

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -285,27 +304,27 @@ abstract class AbstractCheckTask(
     protected open fun completeEvent(canceled: Boolean) {}
 
     /**
-     * Publishes event to the [eventStoreStub].
+     * Publishes event to the [eventBatchRouter].
```
jean-michel-perli-sonsin, line 307:
```suggestion
     * Publishes the event to [eventBatchRouter].
```

src/main/kotlin/com/exactpro/th2/check1/rule/check/CheckRuleTask.kt
```diff
@@ -34,15 +34,16 @@ import java.time.Instant
  * This rule checks the presents of the single message in the messages stream.
```
jean-michel-perli-sonsin, line 34:
```suggestion
 * This rule checks for the presence of a single message in the messages stream.
```

src/main/kotlin/com/exactpro/th2/check1/rule/sequence/SequenceCheckRuleTask.kt
```diff
@@ -60,17 +59,18 @@ import kotlin.collections.set
  * If this parameter is `false` the order won't be checked.
```
jean-michel-perli-sonsin, line 59:
```suggestion
 * If this parameter is set to `false`, the order won't be checked.
```

jean-michel-perli-sonsin (changes requested):
Hello, please see if you agree with the changes suggested below:
README.md
Line 9 - The buffer size is configurable and **it is set to** 1000 by default.

Line 11 - When the component starts, the grpc server **also starts** and then the component waits for incoming grpc requests for verification.

Line 15 - Available requests **are described** in [this repository](https://gitlab.exactpro.com/vivarium/th2/th2-core-open-source/th2-grpc-check1)

Line 17 - CheckSequenceRuleRequest - prefilters **the** messages and verify all of them by **the selected** filter. Order checking **is configured** from request.

Line 18 - CheckRuleRequest - get **the** message filter from request and **compare** it with messages in the cache or await specified time in case of **an empty** cache or message absence.

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
Line 78/86 - Used **to observe** messages in one thread.

Line 79/87 - It provides **a feature** to write **non thread-safe code** in children classes.

Line 116/124 - **Shutdown** the executor that is used to perform this task.

Line 118/126 - throws IllegalStateException if this task has **a connected** task

Lines 128/136 - Registers a task as the next task in **the** continuous verification chain. Its [begin] method will be called

Line 129/137 - when **the current task** completes **a check** or **if a timeout is reached**.

Line 130/138 - The scheduler for **the** current task will be passed to the next task.

Line 132/140 - This method should be called only once**, or otherwise** it throws IllegalStateException.

Line 133/141 - throws IllegalStateException when **the** method is called more than **one time.**

Line 151/159 - Observe a message sequence from **the** checkpoint.

Line 152/160 - **The task subscribes** to messages stream with **its** sequence after **calling it**.

Line 153/161 - This method should be called only once**, or otherwise** it throws IllegalStateException.

Line 155/163 - throws IllegalStateException when **the** method is called more than **one time.**

Line 162/170 - It is called when **the timeout is reached** and **the task is not completed yet**

Line 167/175 - Marks the task as successfully completed. If **the** task timeout had been reached** and then the task was marked as successfully completed**,**

Line 168/177 - the task will be considered as successfully completed**,** because it **was** actually **marked as such.**

Line 185/183 - Provides **a** feature to define **the** custom filter for observe.

Line 190/198 - Observe a message sequence from **the** previous task.

Line 191/199 - **The task subscribes** to messages stream with **its** sequence after **calling it**.

Line 192/200 - This method should be called only once**, or otherwise** it throws IllegalStateException.

Line 193/201 - param sequence**: the** message sequence from **the** previous task.

Line 195/203 - throws IllegalStateException when **the** method is called more than **one time.**

Line 208/216 - All **the** sources above will be disposed on this scheduler.

Line 214/222 - If we move [Observable#unsubscribeOn] after them**,** they won't be disposed until **the** scheduler is free. 

Line 261/280 - Disposes **the** task when timeout is **reached** or **when the** message stream is completed normally or with an exception.

Line 262/281 - **Unsubscribe the task** from **the**message stream.

Line 429/447 - Filters **the** incoming {@link StreamContainer} via session alias and then

Line 430/448 - **it filters the message** which sequence **is greater** than passed

src/main/kotlin/com/exactpro/th2/check1/rule/sequence/SequenceCheckRuleTask.kt
Line 57/56 - If protoFilter parameter is specified**,** the messages will be pre-filtered before going to the actual comparison.

Line 59/58 - If checkOrder parameter is set to `true`**,** the messages must be received in the exact same order as **the** filters **that** were specified.

Line 178/178 - // rule has found **a** complete match for all **the** filters or each filter has found a match by key fields at least

If you have some questions about any comment above, please let me know.
Thanks!

src/main/kotlin/com/exactpro/th2/check1/rule/CheckTaskUtils.kt
```diff
@@ -0,0 +1,101 @@
...
+    val contentSize = builder.getContentSize()
+    check(contentSize <= maxEventBatchContentSize) {
+        "The smallest batch size exceeds the max event batch content size, max $maxEventBatchContentSize, actual $contentSize"
+    }
+
+    return builder.build()
+}
+
+private fun checkAndRebuild(maxEventBatchContentSize: Int, event: ProtoEvent): ProtoEvent {
+    return if (event.body.size() > maxEventBatchContentSize) {
```
OptimumOpium, line 85: I think we need to use `ProtoEvent.getContentSize()` for consistency. It might confuse when we use body.size() in one place and getContentSize() in another

OptimumOpium (approved):

README.md
Nikita-Smirnov-Exactpro: I think the `The Check1 component has two types of pin:` sentence will be more applicable for this case.

OptimumOpium (dismissed):

src/test/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTaskTest.kt
```diff
@@ -18,31 +18,25 @@ import com.exactpro.th2.common.grpc.Direction
...
 abstract class AbstractCheckTaskTest {
     protected val clientStub: MessageRouter<EventBatch> = spy { }
 
-    fun awaitEventBatchRequest(timeoutValue: Long = 1000L): EventBatch {
+    fun awaitEventBatchRequest(timeoutValue: Long = 1000L, times: Int): List<EventBatch> {
         val argumentCaptor = argumentCaptor<EventBatch>()
-        verify(clientStub, timeout(timeoutValue).times(1)).send(argumentCaptor.capture(), any())
-        return assertNotNull(argumentCaptor.lastValue, "Request was not received during $timeoutValue millis")
+        verify(clientStub, timeout(timeoutValue).times(times)).send(argumentCaptor.capture())
+        return assertNotNull(argumentCaptor.allValues, "Request was not received during $timeoutValue millis")
```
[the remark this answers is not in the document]
Nikita-Smirnov-Exactpro: The NotNull is redundant because the previous check already verify count

OptimumOpium (approved):

PR #21, opened 2021-05-24
Th2 1913

Nikita-Smirnov-Exactpro:


jean-michel-perli-sonsin (changes requested):
Hello, please see if you agree with the changes suggested below:

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt 

Line89/90 - It provides **the** feature to write no thread-safe code in children classes.

Line 126/127 - throws IllegalStateException if this task has **a** connected task

Line 137/138 - when **the** current task completes **its** check **or if** timeout is over for it.

Line 138/139 - The scheduler for **the** current task will be passed to the next task.

Line 159/161 - Observe a message sequence from **the** checkpoint.

Line 160/162 - Task subscribe to messages stream with **its** sequence after call.

Line 170/172 -  It is called when **the** timeout is over and **the** task **is not** complete yet

Line 175/177 - **It marks** the task as successfully completed. If **the** task timeout had been exited and then the task was marked as successfully completed

Line 176/178 - the task will be considered as successfully completed **,** because it had actually found that it should

Line 193/195 - Provides **a** feature to define custom filter for observe.

Line 198/200 - Observe a message sequence from **the** previous task.

Line 201/203 - sequence message sequence from **the** previous task.

Line 219/222 - to the actual dispose **that** you want to execute on this scheduler **,**

Line 222/225 - If we move [Observable#unsubscribeOn] after them **,** they won't be disposed until **the** scheduler is free.

Line 280/282 - Disposes **the** task when **the** timeout is over or **the** message stream is completed normally or with an exception.     

Line 281/283 - Task unsubscribe from **the** message stream.

Line 283/285 - @param reason the cause why **a** task must be stopped

Line 448/464 - filters **the** message which **its** sequence **is greater** than passed

In case of any questions, please let me know.

Thanks!

src/main/kotlin/com/exactpro/th2/check1/CollectorService.kt
```diff
@@ -185,11 +185,15 @@ class CollectorService(
         for ((key, task) in eventIdToLastCheckTask.entries) {
```
ivandruzhinin, line 185:
```suggestion
        for ((key, task) in eventIdToLastCheckTask) {
```

src/main/kotlin/com/exactpro/th2/check1/CollectorService.kt
```diff
@@ -185,11 +185,15 @@ class CollectorService(
...
+                runCatching { eventIdToLastCheckTask.compute(key) { _, value ->
+                    if (task == value) {
+                        value.shutdownExecutor()
+                        null
+                    } else {
+                        value
+                    }
+                }}.onFailure { logger.error("Task ${task.description} can't be removed because it has continuation ") }
+                    .onSuccess { logger.debug("Removed task ${task.description} ($endTime) from tasks map") }
             }
```
ivandruzhinin, line 197:
```
if (olderThan(now, delta, unit, endTime) && eventIdToLastCheckTask.remove(key, task)) {
    task.runCatching { shutdownExecutor() }
        .onFailure { logger.error("Task ${task.description} can't be removed because it has continuation ") }
        .onSuccess { logger.debug("Removed task ${task.description} ($endTime) from tasks map") }
}
```

src/test/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTaskTest.kt
```diff
@@ -36,9 +34,28 @@ abstract class AbstractCheckTaskTest {
...
+
+    fun constructMessage(sequence: Long = 0, alias: String = SESSION_ALIAS, type: String = MESSAGE_TYPE, direction: Direction = FIRST): Message.Builder = Message.newBuilder().apply {
+        metadataBuilder.apply {
+            messageType = type
+            idBuilder.apply {
+                this.sequence = sequence
+                this.direction = direction
+                connectionIdBuilder.apply {
+                    this.sessionAlias = alias
+                }
```
ivandruzhinin, line 51:
```suggestion
                connectionIdBuilder.sessionAlias = alias
```

src/test/kotlin/com/exactpro/th2/check1/rule/TestChain.kt
```diff
@@ -0,0 +1,199 @@
...
+        createMessageFilter(sequence),
+        parentEventID,
+        messageStream,
+        clientStub
+    )
+
+    private fun createMessage(sequence: Int) = constructMessage(sequence.toLong())
+        .putAllFields(mapOf(
+            KEY_FIELD to Value.newBuilder().setSimpleValue("$KEY_FIELD$sequence").build(),
+            NOT_KEY_FIELD to Value.newBuilder().setSimpleValue("$NOT_KEY_FIELD$sequence").build()
```
ivandruzhinin, line 182:
```suggestion
            KEY_FIELD to "$KEY_FIELD$sequence".toValue(),
            NOT_KEY_FIELD to "$NOT_KEY_FIELD$sequence".toValue()
```

src/main/kotlin/com/exactpro/th2/check1/CollectorService.kt
ivandruzhinin: Bump

ivandruzhinin: Another approach would be to use `removeIf` on map's entry-set - it would still be easier to understand than removing entry via `compute` method

jean-michel-perli-sonsin (approved):
Hello, please see if you agree with the change below:

src/main/kotlin/com/exactpro/th2/check1/StreamContainer.kt
Line 2 - Copyright **2021-2021** Exactpro (Exactpro Systems Limited)

The rest of the changes were performed accordingly.

Thanks!

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -121,43 +122,46 @@ abstract class AbstractCheckTask(
     }
 
     /**
-     * Shutdowns the executor that is used to perform this task.
-     *
-     * @throws IllegalStateException if this task has connected task
+     * Shutdowns the executor that is used to perform this task if it hasn't got a next task
```
jean-michel-perli-sonsin, line 125:
```suggestion
     * Shutdown the executor that is used to perform this task in case it doesn't have a next task
```

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -121,43 +122,46 @@ abstract class AbstractCheckTask(
...
+     * @return true if the task hasn't got a next task otherwise false
```
jean-michel-perli-sonsin, line 126:
```suggestion
     * @return true if the task doesn't have a next task, otherwise it will return false
```

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -121,43 +122,46 @@ abstract class AbstractCheckTask(
...
      */
-    fun shutdownExecutor() {
+    fun tryShutdownExecutor(): Boolean {
         if (hasNextTask.get()) {
-            throw IllegalStateException("Cannot shutdown executor for task '$description' that has connected task")
+            LOGGER.warn("Cannot shutdown executor for task '$description' that has connected task")
```
jean-michel-perli-sonsin, line 130:
```suggestion
            LOGGER.warn("Cannot shutdown executor for task '$description' that has a connected task")
```

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -121,43 +122,46 @@ abstract class AbstractCheckTask(
...
+            return false
         }
         executorService.shutdown()
+        return true
     }
 
     /**
      * Registers a task as the next task in continuous verification chain. Its [begin] method will be called
```
jean-michel-perli-sonsin, line 138:
```suggestion
     * Registers a task as the next task in the continuous verification chain. Its [begin] method will be called
```

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -121,43 +122,46 @@ abstract class AbstractCheckTask(
...
-     * when current task completes check or timeout is over for it.
-     * The scheduler for current task will be passed to the next task.
+     * when the current task completes its check or if timeout is over for it.
```
jean-michel-perli-sonsin, line 139:
```suggestion
     * when the current task completes its check or if the timeout is over for it.
```

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -445,7 +463,7 @@ abstract class AbstractCheckTask(
 
     /**
      * Filters incoming {@link StreamContainer} via session alias and then
-     * filters message which sequence grete than passed
+     * filters the message whCich its sequence is greater than passed
```
jean-michel-perli-sonsin, line 466:
```suggestion
     * filters the message which its sequence is greater than passed
```

jean-michel-perli-sonsin (approved):
Hi, please just fix some of the comments that I marked.

The rest of the changes are fine.

Thanks!

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -462,7 +463,7 @@ abstract class AbstractCheckTask(
...
```
ivandruzhinin, line 466:
```suggestion
     * filters the message which sequence is greater than a passed one
```

ivandruzhinin (dismissed):

ivandruzhinin (dismissed):

jean-michel-perli-sonsin (dismissed):
All the changes were performed according to the previous comments.
Thanks!

README.md
```diff
@@ -108,6 +108,12 @@ spec:
 
 ## Release Notes
 
+### 2.9.0
+
++ moves the message cursor in the message stream when a filter matched by key fields
```
jean-michel-perli-sonsin, line 113:
```suggestion
+ moves the message cursor in the message stream when a filter is matched by the key fields
```

jean-michel-perli-sonsin (changes requested):
Hi, please see if you agree with the changes suggested below:
README.md

Line 79: The value will be interpreted as time unit defined in _cleanup-time-unit_ setting. _The default value is **set to** 60_

Line 85: The max size in bytes of summary events content in a batch defined in _max-event-batch-content-size_ setting. _The default value is **set to** 1048576_

Line 90: gRPC server pin**: it allow** other components to connect via `com.exactpro.th2.check1.grpc.Check1Service` class.

Line 91: MQ pin**: it is used for listening** to parsed messages. You can link several sources with different directions and session alises to it.

Thanks!

jean-michel-perli-sonsin (approved):
All changes were performed according to the comments.
Thanks!

PR #23, opened 2021-05-28
Th2 1953

Nikita-Smirnov-Exactpro:


README.md
```diff
@@ -108,6 +108,10 @@ spec:
 
 ## Release Notes
 
+### 2.9.1
+
++ fixed problem with using full verification instead of verification by key fields to matching filter in sequence rule when order check is disabled
```
jean-michel-perli-sonsin, line 113:
```suggestion
+ fixed a problem while using full verification instead of verification by key fields, to matching the filter in sequence rule when order check is disabled
```

src/main/kotlin/com/exactpro/th2/check1/rule/sequence/SequenceCheckRuleTask.kt
```diff
@@ -86,7 +84,11 @@ class SequenceCheckRuleTask(
     }
     private lateinit var preFilteringResults: MutableMap<MessageID, ComparisonContainer>
 
+    /**
+     * List of no matched filter yet. It is created from the request filters and reduced after every match
```
jean-michel-perli-sonsin, line 88:
```suggestion
     * List of no matched filter was not created yet. It is created from the requested filters and reduced after every match
```

src/test/kotlin/com/exactpro/th2/check1/rule/sequence/TestSequenceCheckTask.kt
```diff
@@ -379,7 +441,7 @@ class TestSequenceCheckTask : AbstractCheckTaskTest() {
             assertEquals(4, rootEvent.attachedMessageIdsCount) // 3 match key + 1 that doesn't match but between others
```
jean-michel-perli-sonsin, line 441:
```suggestion
            assertEquals(4, rootEvent.attachedMessageIdsCount) // 3 match key + 1 that doesn't match but is between others
```

jean-michel-perli-sonsin (changes requested):
Hello, please see if you agree with the changes proposed below:

src/test/kotlin/com/exactpro/th2/check1/rule/sequence/TestSequenceCheckTask.kt
Line 182/184 - The messages are reordered but **every verification is presented**. So, **every verification** should be passed.

Line 342/404 - goes to processed messages but should not go to **the** actual comparison.

In case of any question, please let me know.
Thanks!

README.md
ivandruzhinin: Why `matching the filter`? I think that `matching filter` here refers to a filter which is used for matching. If that so I think that `to matching filter in` should be changed to `in matching-filter of`

ivandruzhinin (approved):

README.md
jean-michel-perli-sonsin: If "matching filter" refers to a specific filter, then I agree with you that should be changed to **in matching filter of**.

jean-michel-perli-sonsin (dismissed):
All the changes were performed according to the previous comments.
Thanks!

ivandruzhinin (approved):

PR #35, opened 2021-08-09
[TH2-1598] Add functional for new FilterOperations

Sergey-Sitnikov-Exactpro:


src/main/java/com/exactpro/th2/check1/event/VerificationEntryUtils.java
```diff
@@ -55,6 +56,9 @@ public static VerificationEntry createVerificationEntry(ComparisonResult result)
 
     private static String resolveOperation(ComparisonResult result) {
         Object expected = result.getExpected();
+        if (expected instanceof IOperationFilter) {
+            return ((IOperationFilter)expected).getOperation().name();
+        }
```
OptimumOpium, line 61: We should move this statement into the `if (expected instanceof IFilter)` block because IOperationFilter extends IFilter

gradle.properties
```diff
@@ -1,4 +1,4 @@
-release_version = 3.6.1
+release_version = 3.6.2
```
OptimumOpium, line 1: Please, update minor version instead

OptimumOpium (changes requested):

OptimumOpium (approved):

PR #36, opened 2021-08-09
[TH2-1950] Implemented new NoMessageCheck rule task

yarchiksmith:


src/main/kotlin/com/exactpro/th2/check1/CollectorService.kt
```diff
@@ -97,7 +86,8 @@ class CollectorService(
 
         val chainID = request.getChainIdOrGenerate()
 
-        val task = CheckRuleTask(request.description, Instant.now(), SessionKey(sessionAlias, direction), request.timeout, maxEventBatchContentSize,
+        val task = CheckRuleTask(request.description, Instant.now(), SessionKey(sessionAlias, direction),
+            request.messageTimeout, request.timeout, maxEventBatchContentSize,
```
OptimumOpium, line 90: Maybe group messageTimeout and timeout into an object and pass it as a parameter instead of two separate values?

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -235,8 +236,10 @@ abstract class AbstractCheckTask(
             .doOnNext {
                 handledMessageCounter++
 
-                with(it.metadata.id) {
-                    rootEvent.messageID(this)
+                with(it.metadata) {
+                    rootEvent.messageID(this.id)
+                    if (checkOnMessageTimeout(this.timestamp.toInstant()))
```
OptimumOpium, line 241: Please, add brackets to `if` expression

src/main/kotlin/com/exactpro/th2/check1/rule/sequence/NoMessageCheckTask.kt
```diff
@@ -0,0 +1,134 @@
...
+        resultEvent = Event.start()
+            .type("noMessagesCheckResult")
+        rootEvent.addSubEvent(resultEvent)
+    }
+
+    override fun onNext(messageContainer: MessageContainer) {
+        LOGGER.debug(
+            "Received message with id: {}",
+            TextFormat.shortDebugString(messageContainer.protoMessage.metadata.id)
+        )
```
OptimumOpium, line 91: You should wrap the call in `if` condition and call it only if the debug level is enabled

src/main/kotlin/com/exactpro/th2/check1/rule/sequence/NoMessageCheckTask.kt
```diff
@@ -0,0 +1,134 @@
...
+        } else {
+            extraMessagesCounter++
+            resultEvent.messageID(comparisonContainer.protoActual.metadata.id)
+        }
+    }
+
+    override fun completeEvent(canceled: Boolean) {
+        preFilterEvent.name("Prefilter: $preFilterMessagesCounter messages were filtered.")
+
+        if (extraMessagesCounter == 0)
```
OptimumOpium, line 112: Please, add brackets to the `if-else` statement

src/main/kotlin/com/exactpro/th2/check1/utils/TimeUtils.kt
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+
+package com.exactpro.th2.check1.utils
+
+import com.google.protobuf.Timestamp
+import java.time.Instant
+
+fun Timestamp.toInstant(): Instant = Instant.ofEpochSecond(seconds, nanos.toLong())
+fun Instant.toTimestamp(): Timestamp = Timestamp.newBuilder().setSeconds(epochSecond).setNanos(nano).build()
```
OptimumOpium, line 20: Ther is a method in MessageUtils that does the same

src/main/kotlin/com/exactpro/th2/check1/rule/sequence/NoMessageCheckTask.kt
```diff
@@ -0,0 +1,134 @@
...
+        .also {
+            if (hasMetadataFilter()) {
+                it.metadataFilter = metadataFilter
+            }
+        }
+        .build()
+
+    private fun PreFilter.toMessageFilter() = MessageFilter.newBuilder()
+        .putAllFields(fieldsMap)
+        .build()
```
OptimumOpium, line 133: The same methods are used in CheckSequenceRule. We probably should extract those methods into a utility class or into the AbstractCheckTask and reuse them

OptimumOpium (changes requested):
Please, add tests for a new rule

src/main/java/com/exactpro/th2/check1/Checkpoint.java
```diff
@@ -97,27 +98,35 @@ public boolean equals(Object obj) {
...
+        Map<SessionKey, CheckpointData> sessionKeyToSequence = new HashMap<>();
         for (Map.Entry<String, DirectionCheckpoint> sessionAliasDirectionCheckpointEntry : protoCheckpoint.getSessionAliasToDirectionCheckpointMap().entrySet()) {
             String sessionAlias = sessionAliasDirectionCheckpointEntry.getKey();
             DirectionCheckpoint directionCheckpoint = sessionAliasDirectionCheckpointEntry.getValue();
-            for (Map.Entry<Integer, Long> directionSequenceEntry : directionCheckpoint.getDirectionToSequenceMap().entrySet()) {
-                SessionKey sessionKey = new SessionKey(sessionAlias, Direction.forNumber(directionSequenceEntry.getKey()));
-                sessionKeyToSequence.put(sessionKey, directionSequenceEntry.getValue());
-            }
+            if (directionCheckpoint.getDirectionToCheckpointDataCount() != 0) {
+                for (Map.Entry<Integer, com.exactpro.th2.common.grpc.Checkpoint.CheckpointData> directionSequenceEntry : directionCheckpoint.getDirectionToCheckpointDataMap().entrySet()) {
```
OptimumOpium, line 119: Why don't you check if the gRPC checkpoint has only one map set? We should report an error if both old and new version are used

src/main/kotlin/com/exactpro/th2/check1/rule/sequence/NoMessageCheckTask.kt
```diff
@@ -0,0 +1,117 @@
...
+
+        val result = matchFilter(
+            messageContainer,
+            messagePreFilter,
+            metadataPreFilter,
+            matchNames = false,
+            significant = false
+        )
+        val comparisonContainer = ComparisonContainer(messageContainer, protoPreMessageFilter, result)
+        if (comparisonContainer.fullyMatches) {
```
OptimumOpium, line 98: If you only want to use `fullyMatches` you probably should move the extension method for AggregatedResult method from ComparisonContainer to some utility class. No needs to create an object just to access one method

OptimumOpium (changes requested):

src/main/kotlin/com/exactpro/th2/check1/entities/Checkpoint.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import com.datastax.driver.core.utils.UUIDs.timeBased
+import com.exactpro.th2.check1.SessionKey
+
+data class Checkpoint(
+    val id: String = timeBased().toString(), val sessionKeyToCheckpointData: Map<SessionKey, CheckpointData>
+) {
+    constructor(sessionKeyToCheckpointData: Map<SessionKey, CheckpointData>) : this(
+        timeBased().toString(), sessionKeyToCheckpointData
+    )
```
OptimumOpium, line 11: There is no need to have a secondary constructor. You can add @JvmOverlods annotation on the primary one. By the way, why do you need a secondary constructor like this? I did not find any place where you use it

src/main/kotlin/com/exactpro/th2/check1/entities/CheckpointData.kt
```diff
@@ -0,0 +1,5 @@
+package com.exactpro.th2.check1.entities
```
OptimumOpium, line 1: Missing copyright

src/main/kotlin/com/exactpro/th2/check1/entities/TaskTimeout.kt
```diff
@@ -0,0 +1,5 @@
+package com.exactpro.th2.check1.entities
```
OptimumOpium, line 1: Missing copyright

src/main/kotlin/com/exactpro/th2/check1/entities/Checkpoint.kt
```diff
@@ -0,0 +1,12 @@
+package com.exactpro.th2.check1.entities
+
+import com.datastax.driver.core.utils.UUIDs.timeBased
```
OptimumOpium, line 3: Just to avoid using the Cassandra dependency we can use `EventUtils.generateUUID()`. It uses the time-base generation as well

src/main/kotlin/com/exactpro/th2/check1/entities/TaskTimeout.kt
```diff
@@ -0,0 +1,5 @@
...
+
+data class TaskTimeout(val messageTimeout: Long? = null, val timeout: Long) {
```
OptimumOpium, line 3: If you want to use the default values you should put the parameters without ones first

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -214,13 +216,13 @@ abstract class AbstractCheckTask(
         if (!taskState.compareAndSet(State.CREATED, State.BEGIN)) {
             throw IllegalStateException("Task $description already has been started")
         }
-        LOGGER.info("Check begin for session alias '{}' with sequence '{}' timeout '{}'", sessionKey, sequence, timeout)
+        LOGGER.info("Check begin for session alias '{}' with sequence '{}' timeout '{}'", sessionKey, sequence, taskTimeout.timeout)
```
OptimumOpium, line 219: Please, add information about message timeout as well (or you can simply insert the whole **taskTimeout** into the string)

src/main/kotlin/com/exactpro/th2/check1/rule/sequence/NoMessageCheckTask.kt
```diff
@@ -94,13 +95,14 @@ class NoMessageCheckTask(
...
             significant = false
         )
-        val comparisonContainer = ComparisonContainer(messageContainer, protoPreMessageFilter, result)
-        if (comparisonContainer.fullyMatches) {
-            preFilterMessagesCounter++
-            preFilterEvent.messageID(comparisonContainer.protoActual.metadata.id)
-        } else {
-            extraMessagesCounter++
-            resultEvent.messageID(comparisonContainer.protoActual.metadata.id)
+        messageContainer.protoMessage.metadata.run {
```
OptimumOpium, line 98: When you don't want to return anything it is better to use `apply` (or `also`) instead
```suggestion
        messageContainer.protoMessage.metadata.apply {
```

src/main/kotlin/com/exactpro/th2/check1/utils/ProtoMessageUtils.kt
```diff
@@ -22,4 +40,51 @@ fun PreFilter.toRootMessageFilter(): RootMessageFilter = RootMessageFilter.newBu
...
+        checkpointBuilder.putSessionAliasToDirectionCheckpoint(sessionAlias, directionCheckpoint.build())
+    }
+
+    return checkpointBuilder.build()
+}
+
+fun Checkpoint.convert(): com.exactpro.th2.check1.entities.Checkpoint {
+    val sessionKeyToSequence: MutableMap<SessionKey, CheckpointData> = HashMap()
+    sessionAliasToDirectionCheckpointMap.forEach { (sessionAlias, directionCheckpoint) ->
+        check(!(directionCheckpoint.directionToCheckpointDataCount != 0 && directionCheckpoint.directionToSequenceCount != 0)) {
```
OptimumOpium, line 73:
```suggestion
        check(!directionCheckpoint.run { directionToCheckpointDataCount != 0 && directionToSequenceCount != 0 }) {
```

src/main/kotlin/com/exactpro/th2/check1/utils/ProtoMessageUtils.kt
```diff
@@ -22,4 +40,51 @@ fun PreFilter.toRootMessageFilter(): RootMessageFilter = RootMessageFilter.newBu
...
+            }
+        } else {
+            directionCheckpoint.directionToCheckpointDataMap.forEach { (directionNumber, checkpointData) ->
+                val sessionKey = SessionKey(sessionAlias, Direction.forNumber(directionNumber))
+                sessionKeyToSequence[sessionKey] = checkpointData.convert()
+
+            }
+        }
+    }
+    return com.exactpro.th2.check1.entities.Checkpoint(id, sessionKeyToSequence)
```
OptimumOpium, line 89: You can use aliases in Kotlin to avoid fully qualified names

src/test/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTaskTest.kt
```diff
@@ -40,9 +44,16 @@ abstract class AbstractCheckTaskTest {
         )
     }
 
-    fun constructMessage(sequence: Long = 0, alias: String = SESSION_ALIAS, type: String = MESSAGE_TYPE, direction: Direction = FIRST): Message.Builder = Message.newBuilder().apply {
+    fun constructMessage(
+        sequence: Long = 0,
+        alias: String = SESSION_ALIAS,
+        type: String = MESSAGE_TYPE,
+        direction: Direction = FIRST,
+        timestamp: Timestamp? = Timestamp.getDefaultInstance()
```
OptimumOpium, line 52: Why do you need a nullable timestamp for the message? If you using default value for argument there is no need to use nullable type (in most cases)

src/test/kotlin/com/exactpro/th2/check1/rule/sequence/TestNoMessageCheckTask.kt
```diff
@@ -0,0 +1,191 @@
...
+        var sequence = 1L;
+        val messages: MutableList<Message> = ArrayList()
+        messageData.forEach { data ->
+            messages.add(
+                constructMessage(sequence++, sessionAlias, messageType, direction, data.timestamp)
+                    .putFields(data.fieldName, data.value)
+                    .build()
+            )
+        }
+        return messages
```
OptimumOpium, line 160:
```suggestion
        return messageData.map { data ->
                constructMessage(sequence++, sessionAlias, messageType, direction, data.timestamp)
                    .putFields(data.fieldName, data.value)
                    .build()
        }
```

src/test/kotlin/com/exactpro/th2/check1/rule/sequence/TestNoMessageCheckTask.kt
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    private fun createValue(value: String): Value = Value.newBuilder().setSimpleValue(value).build()
```
OptimumOpium, line 163: There is an extension function `Any.toValue()` that can be used

OptimumOpium (changes requested):

src/main/kotlin/com/exactpro/th2/check1/utils/ProtoMessageUtils.kt
```diff
@@ -24,6 +23,9 @@ import com.exactpro.th2.common.grpc.Direction
 import com.exactpro.th2.common.grpc.MessageFilter
 import com.exactpro.th2.common.grpc.RootMessageFilter
 import com.exactpro.th2.sailfish.utils.ProtoToIMessageConverter
+import com.exactpro.th2.check1.entities.Checkpoint as InternalCheckpoint
+import com.exactpro.th2.check1.entities.CheckpointData as InternalCheckpointData
+import com.exactpro.th2.common.grpc.Checkpoint.CheckpointData as CheckpointData
```
OptimumOpium, line 28: There is no need to use aliases for inner classes. They can be simple imported
```suggestion
import com.exactpro.th2.common.grpc.Checkpoint.CheckpointData
```

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -506,6 +520,18 @@ abstract class AbstractCheckTask(
...
 
+    /**
+     * Provides the ability to stop observing if a message timeout is set.
+     */
+    private fun Observable<Message>.onMessageTimeout() : Observable<Message> =
+        takeWhile {
+            checkOnMessageTimeout(it.metadata.timestamp).also { continueObservation ->
+                if (!continueObservation) {
+                    end(State.MESSAGE_TIMEOUT, "Expired message timeout")
+                }
```
OptimumOpium, line 531: I believe we should not call the `end` method right here. We should do that only in terminated methods: onError, onComplete, the timeout is exited in timeout subscriber.

We can set a volatile variable with the status that should be passed when `onComplete` method will be called. By default, that status will be STREAM_COMPLETED but if this method is triggered the status will be changed to MESSAGE_TIMEOUT

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -506,6 +520,18 @@ abstract class AbstractCheckTask(
         return checkpointTimeout == null || checkpointTimeout!!.isAfter(timestamp) || checkpointTimeout == timestamp
     }
 
+    /**
+     * Provides the ability to stop observing if a message timeout is set.
+     */
+    private fun Observable<Message>.onMessageTimeout() : Observable<Message> =
```
OptimumOpium, line 526: Let's rename it into `takeWhileMessagesInTimeout`

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -255,7 +268,7 @@ abstract class AbstractCheckTask(
         try {
             val currentState = taskState.get()
             LOGGER.info("Finishes task '$description' in state $currentState")
-            if (currentState == State.TIMEOUT) {
+            if (currentState == State.TIMEOUT || currentState == State.MESSAGE_TIMEOUT || currentState == State.STREAM_COMPLETED) {
```
OptimumOpium, line 271: I believe we can add the field into the enum that indicated whether the state should trigger `callOnTimeoutCallback` method call. And it will simplify the condition

src/main/kotlin/com/exactpro/th2/check1/rule/nomessage/NoMessageCheckTask.kt
```diff
@@ -115,5 +115,7 @@ class NoMessageCheckTask(
             resultEvent.status(Event.Status.FAILED)
                 .name("Check failed: $extraMessagesCounter extra messages were found.")
         }
+
+        resultEvent.description("Task has been completed because: {$taskState}")
```
OptimumOpium, line 119: Maybe we should fail the check rule if the taskState == TIMEOUT or STREAM_COMPLETED because in these cases we cannot be sure if there really was not any message during the timeout the user wanted.

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -299,22 +342,24 @@ abstract class AbstractCheckTask(
...
+    protected open fun completeEvent(taskState: State) {}
 
     /**
      * Publishes the event to [eventBatchRouter].
      */
     private fun publishEvent() {
         val prevState = taskState.getAndSet(State.PUBLISHED)
         if (prevState != State.PUBLISHED) {
-            completeEvent(prevState == State.TIMEOUT)
+            if (!untrusted || isIgnoreUntrustedFlag()) {
```
OptimumOpium, line 360: Why do we ignore the untrusted flag in some cases?

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -337,6 +382,34 @@ abstract class AbstractCheckTask(
         }
     }
 
+    private fun doOnCompleteEvent(previousState: State) {
+        completeEvent(previousState)
```
OptimumOpium, line 386: Why do we call `completeEvent` here? It should be called regardless of the  `untrusted` flag

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -204,20 +236,28 @@ abstract class AbstractCheckTask(
...
         if (!taskState.compareAndSet(State.CREATED, State.BEGIN)) {
             throw IllegalStateException("Task $description already has been started")
         }
-        LOGGER.info("Check begin for session alias '{}' with sequence '{}' timeout '{}'", sessionKey, sequence, timeout)
+        LOGGER.info("Check begin for session alias '{}' with sequence '{}' and task timeout '{}'", sessionKey, sequence, taskTimeout)
         this.lastSequence = sequence
         this.executorService = executorService
+        if (untrusted && !isIgnoreUntrustedFlag()) {
+            doOnUntrustedExecution()
+            return
```
OptimumOpium, line 254: Why don't we start the task when the untrusted checkpoint is passed?

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -267,7 +308,8 @@ abstract class AbstractCheckTask(
                     .toProto(parentEventID))
                 .build())
         } finally {
-            sequenceSubject.onSuccess(Legacy(executorService, lastSequence))
+            val untrustedForFutureTask = !bufferContainsStartMessage && !hasMessagesInTimeoutInterval
```
OptimumOpium, line 311: Maybe `!hasMessagesInTimeoutInterval` will be enough here

src/main/kotlin/com/exactpro/th2/check1/utils/TimeUtils.kt
```diff
@@ -0,0 +1,31 @@
...
+        return 1
+    if (this.nanos < other.nanos)
+        return -1
+    if (this.nanos > other.nanos)
+        return 1
+    return 0
+}
+
+fun Timestamp.isBefore(other: Timestamp): Boolean = this < other
+fun Timestamp.isAfter(other: Timestamp): Boolean = this > other
```
OptimumOpium, line 31:
```suggestion
private val timestampComparator =         Comparator.comparingLong(Timestamp::getSeconds).thenComparingInt(Timestamp::getNanos)
operator fun Timestamp.compareTo(other: Timestamp): Int = timestampComparator.compare(this, other)

fun Timestamp.isBefore(other: Timestamp): Boolean = this < other
fun Timestamp.isAfter(other: Timestamp): Boolean = this > other
```

src/test/kotlin/com/exactpro/th2/check1/rule/TestChain.kt
```diff
@@ -70,9 +74,9 @@ class TestChain: AbstractCheckTaskTest() {
 
     @Test
     fun `sequence rules - two succeed`() {
-        val streams = createStreams(messages = (1..4).map(::createMessage))
+        val streams = createStreams(messages = (0..4).map(::createMessage))
```
OptimumOpium, line 77: Why have you changed the sequence?

OptimumOpium (changes requested):

src/test/kotlin/com/exactpro/th2/check1/rule/TestChain.kt
yarchiksmith: Because we will receive an additional event with the failed status, which will notify about the missed start message in the stream

OptimumOpium: The checkpoint does not work properly then. We don't pass the checkpoint here at all. So why does it reports about missing start message if we did not set one?

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
yarchiksmith: But shouldn't we check that there are no messages in the specified interval before we passed untrusted flag?

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
yarchiksmith: If we passed untrusted flag, some pre initialized data in method `onStart` will be ignored due to `onSubscribe` method not being called, so we have to check that

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
OptimumOpium: If the `!hasMessagesInTimeoutInterval` then we didn't find any message that has a timestamp less than the timestamp in the checkpoint. If the message from the checkpoint was in the buffer we would found it and the `hasMessagesInTimeoutInterval` would be `true`. If I am not mistaking of course

yarchiksmith:
@OptimumOpium So if we have no messages in the interval or a missed start message from the previous task, should we still need to execute the task? In other words, is there no reason not to execute the task?

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
yarchiksmith: Refer to https://github.com/th2-net/th2-check1/pull/36#issuecomment-899487957

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
yarchiksmith: Refer to https://github.com/th2-net/th2-check1/pull/36#issuecomment-899487957

yarchiksmith: Refer to https://github.com/th2-net/th2-check1/pull/36#issuecomment-899487957

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
yarchiksmith: Message timestamp timeout is checkpoint timestamp + message timeout. If the buffer missed the checkpoint message and the next message following it that falls in this interval, then we will get `bufferContainsStartMessage = false` and `hasMessagesInTimeoutInterval = true`

OptimumOpium: > @OptimumOpium So if we have no messages in the interval or a missed start message from the previous task, should we still need to execute the task? In other words, is there no reason not to execute the task?

In my opinion, yes. We should execute the check rule event if it is started from an untrusted checkpoint. If you have doubts about that we can add a parameter that will configure the check1 to behave the way we need (fail the rule if it is started from the untrusted checkpoint without any execution or try to execute the next rule and add the failed event to the result regardless the rule succeeds or not).

OptimumOpium:
> @OptimumOpium So if we have no messages in the interval or a missed start message from the previous task, should we still need to execute the task? In other words, is there no reason not to execute the task?

In my opinion, yes. We should execute the check rule event if it is started from an untrusted checkpoint. If you have doubts about that we can add a parameter that will configure the check1 to behave the way we need (fail the rule if it is started from the untrusted checkpoint without any execution or try to execute the next rule and add the failed event to the result regardless the rule succeeds or not).

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
OptimumOpium: And what? In this case, the checkpoint is not untrusted and the `!hasMessagesInTimeoutInterval` will return `false` (as it should)

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -152,7 +171,9 @@ abstract class AbstractCheckTask(
                     } else {
                         legacy.executorService
                     }
-                checkTask.begin(legacy.lastSequence, executor)
+                legacy.sequenceData.apply {
+                    checkTask.begin(this.lastSequence, this.lastMessageTimestamp, executor, this.untrusted)
```
OptimumOpium, line 175: You don't need to use `this` here. Because of `apply` the correct context should be chosen

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -337,6 +377,42 @@ abstract class AbstractCheckTask(
...
+        )
+    }
+
+    private fun fillEmptyStarMessageEvent() {
+        rootEvent.addSubEvent(
+            Event.start()
+                .name("Buffer for session alias '${sessionKey.sessionAlias}' and direction '${sessionKey.direction}' doesn't contain starting message, but contains several messages in the requested check interval")
+                .status(FAILED)
+        )
+    }
```
OptimumOpium, line 414: Please, add type for events

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -337,6 +377,42 @@ abstract class AbstractCheckTask(
         }
     }
 
+    private fun doAfterCompleteEvent() {
+        if (untrusted && !isIgnoreUntrustedFlag()) {
```
OptimumOpium, line 381: Why do we still have `isIgnoreUntrustedFlag()` method? I don't understand why do we need it. Could you please explain this to me one more time?

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -469,22 +550,61 @@ abstract class AbstractCheckTask(
     private fun Observable<StreamContainer>.continueObserve(sessionKey: SessionKey, sequence: Long): Observable<Message> =
         filter { streamContainer -> streamContainer.sessionKey == sessionKey }
             .flatMap(StreamContainer::bufferedMessages)
-            .filter { message -> message.metadata.id.sequence > sequence }
+            .filter { message ->
+                if (message.metadata.id.sequence == sequence || sequence == DEFAULT_SEQUENCE) {
+                    bufferContainsStartMessage = true
```
OptimumOpium, line 555: I think we can set `bufferContainsStartMessage` to `true` if `sequence == DEFAULT_SEQUENCE` right in the begin method. We can remove this condition from here then

src/main/kotlin/com/exactpro/th2/check1/rule/check/CheckRuleTask.kt
```diff
@@ -92,4 +93,6 @@ class CheckRuleTask(
             .type("Check failed")
             .status(FAILED)
     }
+
+    override fun isIgnoreUntrustedFlag(): Boolean = true
```
OptimumOpium, line 97: Why do we need to ignore the untrusted flag for this rule?

src/main/kotlin/com/exactpro/th2/check1/utils/TimeUtils.kt
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright 2020-2021 Exactpro (Exactpro Systems Limited)
```
OptimumOpium, line 2:
```suggestion
 * Copyright 2021-2021 Exactpro (Exactpro Systems Limited)
```

src/test/kotlin/com/exactpro/th2/check1/rule/nomessage/TestNoMessageCheckTask.kt
```diff
@@ -0,0 +1,184 @@
...
+        )
+        task.begin(createCheckpoint(checkpointTimestamp, 1))
+
+        val eventBatch = awaitEventBatchRequest(1000L, 2)
+        val eventsList = eventBatch.flatMap(EventBatch::getEventsList)
+
+        assertAll({
+            assertTrue(eventsList.all { it.status == EventStatus.SUCCESS }, "Has messages outside the prefilter")
+            assertTrue(eventsList.first().attachedMessageIdsCount == 5)
+            assertTrue(eventsList[1].attachedMessageIdsCount == 4)
```
OptimumOpium, line 71: I would say, that we should also check that the correct IDs are attached

src/test/kotlin/com/exactpro/th2/check1/rule/sequence/TestSequenceCheckTask.kt
```diff
@@ -316,6 +316,222 @@ class TestSequenceCheckTask : AbstractCheckTaskTest() {
...
+            assertEquals(3, verifications.size, "Unexpected verifications count: $verifications")
+            assertTrue("Some verifications are not success: $verifications") { verifications.all { it.status == EventStatus.SUCCESS } }
+            assertEquals(listOf(1L, 2L, 3L), verifications.flatMap { verification -> verification.attachedMessageIdsList.map { it.sequence } })
+        }, {
+            assertCheckSequenceStatus(EventStatus.SUCCESS, eventsList) // because all key fields are in a correct order
+        })
+    }
+
+    @Test
+    fun `check sequence of messages with the same value of key field and expired message timeout`() {
```
OptimumOpium, line 398: What about the case when one of the messages is outside the message timeout?

OptimumOpium (changes requested):

src/main/kotlin/com/exactpro/th2/check1/utils/ProtoMessageUtils.kt
```diff
@@ -0,0 +1,92 @@
...
+    sessionKeyToCheckpointData.forEach { (sessionKey, checkpointData) ->
+        intermediateMap.computeIfAbsent(sessionKey.sessionAlias) {
+            DirectionCheckpoint.newBuilder()
+        }.putDirectionToCheckpointData(sessionKey.direction.number, checkpointData.convert())
+    }
+
+    val checkpointBuilder = Checkpoint.newBuilder().setId(id)
+    intermediateMap.forEach { (sessionAlias, directionCheckpoint) -> 
+        checkpointBuilder.putSessionAliasToDirectionCheckpoint(sessionAlias, directionCheckpoint.build())
+    }
```
OptimumOpium, line 67: We currently breaking backward compatibility because we are setting only the new checkpoint data. But old version of gRPC won't be able to parse this data

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -469,22 +552,61 @@ abstract class AbstractCheckTask(
...
+        }
+
+        return checkpointData.convert().apply {
             LOGGER.info("Use sequence '{}' from checkpoint for session '{}'", sequence, sessionKey)
         }
+    }
 
-        return sequence ?: DEFAULT_SEQUENCE
+    private fun checkOnMessageTimeout(timestamp: Timestamp): Boolean {
+        return checkpointTimeout == null || checkpointTimeout!!.isAfter(timestamp) || checkpointTimeout == timestamp
```
OptimumOpium, line 587: I believe we don't need `!!` because we get there only if the `checkpointTimeout != null`

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -469,22 +552,61 @@ abstract class AbstractCheckTask(
...
+            checkOnMessageTimeout(it.metadata.timestamp).also { continueObservation ->
+                hasMessagesInTimeoutInterval = hasMessagesInTimeoutInterval or continueObservation
+                if (!continueObservation) {
+                    streamCompletedState = State.MESSAGE_TIMEOUT
+                }
+            }
+        }
+
+    private fun calculateCheckpointTimeout(timestamp: Timestamp?, messageTimeout: Long?): Timestamp? =
+        if (timestamp == null || messageTimeout == null) {
```
OptimumOpium, line 604: Shouldn't we alert if the `messageTimeout` is set but the checkpoint does not have `timestamp`?

OptimumOpium (changes requested):

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
yarchiksmith: Kotlin cannot smart cast 'Timestamp?' to 'Timestamp' here

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -204,20 +236,26 @@ abstract class AbstractCheckTask(
...
         if (!taskState.compareAndSet(State.CREATED, State.BEGIN)) {
             throw IllegalStateException("Task $description already has been started")
         }
-        LOGGER.info("Check begin for session alias '{}' with sequence '{}' timeout '{}'", sessionKey, sequence, timeout)
+        LOGGER.info("Check begin for session alias '{}' with sequence '{}' and task timeout '{}'", sessionKey, sequence, taskTimeout)
         this.lastSequence = sequence
         this.executorService = executorService
+        this.untrusted = untrusted
+        this.checkpointTimeout = calculateCheckpointTimeout(checkpointTimestamp, taskTimeout.messageTimeout)
+        this.bufferContainsStartMessage = sequence == DEFAULT_SEQUENCE
```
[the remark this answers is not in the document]
yarchiksmith: Using this condition, we can check that the sequence was passed in the request. If the initial value is set to false, then we will believe that the request always contains a sequence and will try to check it

src/main/kotlin/com/exactpro/th2/check1/rule/nomessage/NoMessageCheckTask.kt
```diff
@@ -0,0 +1,125 @@
...
+            .type("noMessagesCheckResult")
+        rootEvent.addSubEvent(resultEvent)
+    }
+
+    override fun onNext(messageContainer: MessageContainer) {
+        if (LOGGER.isDebugEnabled) {
+            LOGGER.debug("Received message with id: {}", messageContainer.protoMessage.metadata.id.toJson())
+        }
+
+        val result = matchFilter(
```
Nikita-Smirnov-Exactpro, line 91: !! Use the same approach to execute per filtering as in SequenceCheckRuleTask class

Nikita-Smirnov-Exactpro (changes requested):

Nikita-Smirnov-Exactpro (dismissed):

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -608,6 +610,18 @@ abstract class AbstractCheckTask(
...
+    private fun TaskTimeout.getOrCalculateMessageTimeout(): Long {
+        if (this.timeout > 0L) {
+            return timeout
+        }
+        require(messageTimeout != null && messageTimeout > 0) {
+            "Timeout cannot be calculated because 'timeout' and 'message timeout' is not set"
+        }
+        return messageTimeout.also {
+            LOGGER.info("Rule `timeout` is not specified, used `message timeout` instead")
+        }
```
OptimumOpium, line 622: Don't see any reason to use `also` here. You can easily split this
```suggestion
        LOGGER.info("Rule `timeout` is not specified, used `message timeout` instead")
        return messageTimeout
```

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -608,6 +610,18 @@ abstract class AbstractCheckTask(
...
         }
 
+    private fun TaskTimeout.getOrCalculateMessageTimeout(): Long {
+        if (this.timeout > 0L) {
+            return timeout
+        }
+        require(messageTimeout != null && messageTimeout > 0) {
+            "Timeout cannot be calculated because 'timeout' and 'message timeout' is not set"
+        }
+        return messageTimeout.also {
```
OptimumOpium, line 620: I don't think that we can use the `messageTimeout` as a parameter for the timer task to stop our rule. The timer task is executed in real-time. So the `timeout` must be greater that the `messageTimeout`

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -126,6 +126,8 @@ abstract class AbstractCheckTask(
     private var hasMessagesInTimeoutInterval: Boolean = false
     private var bufferContainsStartMessage: Boolean = false
     private var isDefaultSequence: Boolean = false
+    protected var isCheckpointLastReceivedMessage = bufferContainsStartMessage && !hasMessagesInTimeoutInterval
```
OptimumOpium, line 129: I don't see any places where you set this property a new value. So it is always `false`

OptimumOpium (dismissed):

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
yarchiksmith: So we should to add an additional timeout to the message timeout? Something like this `message timeout + 5000 mills`

PR #44, opened 2021-08-23
[TH2-1614] Added logging about received message type in the message filter

yarchiksmith:


src/main/kotlin/com/exactpro/th2/check1/CollectorService.kt
```diff
@@ -141,10 +141,14 @@ class CollectorService(
 
     private fun MessageFilter.toRootMessageFilter(): RootMessageFilter {
         return RootMessageFilter.newBuilder()
-                .setMessageType(this.messageType)
-                .setComparisonSettings(this.comparisonSettings.toRootComparisonSettings())
-                .setMessageFilter(this)
-                .build()
+            .setMessageType(this.messageType).also {
+                if (this.messageType.isNotEmpty()) {
```
OptimumOpium, line 145: I think we should report it as an exception. We cannot continue the rule execution because we don't have a message type to check.
The user should know that the rule's parameters are incorrect as soon as possible

OptimumOpium (changes requested):

src/main/kotlin/com/exactpro/th2/check1/CollectorService.kt
```diff
@@ -141,10 +141,13 @@ class CollectorService(
 
     private fun MessageFilter.toRootMessageFilter(): RootMessageFilter {
         return RootMessageFilter.newBuilder()
-                .setMessageType(this.messageType)
-                .setComparisonSettings(this.comparisonSettings.toRootComparisonSettings())
-                .setMessageFilter(this)
-                .build()
+            .setMessageType(this.messageType).also {
+                check(this.messageType.isNotEmpty()) { "Rule cannot be executed because the message filter does not contain 'message type'" }
```
Nikita-Smirnov-Exactpro, line 145: Please execute check at the start of the method to avoid redundant operations

Nikita-Smirnov-Exactpro (approved):

src/main/kotlin/com/exactpro/th2/check1/CollectorService.kt
```diff
@@ -140,11 +140,14 @@ class CollectorService(
...
 
     private fun MessageFilter.toRootMessageFilter(): RootMessageFilter {
+        check(this.messageType.isNotEmpty()) { "Rule cannot be executed because the message filter does not contain 'message type'" }
         return RootMessageFilter.newBuilder()
-                .setMessageType(this.messageType)
-                .setComparisonSettings(this.comparisonSettings.toRootComparisonSettings())
-                .setMessageFilter(this)
-                .build()
+            .setMessageType(this.messageType).also {
+                logger.info("Rule used a message type '${this.messageType}' from the message filter")
```
OptimumOpium, line 146: Why do we need this logging here? We will be able to see that in the original request

src/main/kotlin/com/exactpro/th2/check1/CollectorService.kt
```diff
@@ -140,11 +140,14 @@ class CollectorService(
     }
 
     private fun MessageFilter.toRootMessageFilter(): RootMessageFilter {
+        check(this.messageType.isNotEmpty()) { "Rule cannot be executed because the message filter does not contain 'message type'" }
```
OptimumOpium, line 143: I would use `isNotBlank` instead of `isNotEmpty`. There is no point of having message with message type consists of whitespaces only

OptimumOpium (dismissed):

src/main/kotlin/com/exactpro/th2/check1/rule/RuleFactory.kt
```diff
@@ -138,6 +138,7 @@ class RuleFactory(
     }
 
     private fun MessageFilter.toRootMessageFilter(): RootMessageFilter {
+        check(this.messageType.isNotBlank()) { "Rule cannot be executed because the message filter does not contain 'message type'" }
```
OptimumOpium, line 141: Shouldn't we check the message type for `RootMessageFilter` as well? The CheckRuleRequest/CheckSequenceRuleRequest might contain the RootMessageFilter instead of MessageFilter

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -503,6 +503,9 @@ abstract class AbstractCheckTask(
 
     protected fun ProtoToIMessageConverter.fromProtoPreFilter(protoPreMessageFilter: RootMessageFilter,
                                                               messageName: String = protoPreMessageFilter.messageType): IMessage {
+        check(messageName.isNotBlank()) {
+            "Rule cannot be created or executed because the root message filter does not contain 'message type'"
+        }
```
OptimumOpium, line 508: I think it is a bit late to check the message type. I think we should do that in the RuleFactory before rule creation. That will allow us to report an error before submitting the rule

src/main/kotlin/com/exactpro/th2/check1/rule/RuleFactory.kt
```diff
@@ -93,8 +93,10 @@ class RuleFactory(
                         "Either messageFilters or rootMessageFilters must be set but not both"
                     }
 
-                    val protoMessageFilters: List<RootMessageFilter> = request.rootMessageFiltersList.ifEmpty {
+                    val protoMessageFilters: List<RootMessageFilter> = if (request.rootMessageFiltersList.isEmpty()) {
                         request.messageFiltersList.map { it.toRootMessageFilter() }
+                    } else {
+                        request.rootMessageFiltersList.onEach { checkMessageType(it.messageType) }
                     }
```
OptimumOpium, line 100:
```suggestion
                    val protoMessageFilters: List<RootMessageFilter> =request.rootMessageFiltersList.ifEmpty {
                        request.messageFiltersList.map { it.toRootMessageFilter() }
                    }.onEach { checkMessageType(it.messageType) }
```

yarchiksmith: If `rootMessageFiltersList` is empty, then `checkMessageType` will be called twice for each element

OptimumOpium: Yep, but calling that method does not cost match for us. We can either remove the `checkMessageType` in `toRootMessageFilter` method or agree that calling `checkMessageType` twice for each filter is okay

src/main/kotlin/com/exactpro/th2/check1/rule/RuleFactory.kt
```diff
@@ -140,16 +138,15 @@ class RuleFactory(
...
         return RootMessageFilter.newBuilder()
                 .setMessageType(this.messageType)
                 .setComparisonSettings(this.comparisonSettings.toRootComparisonSettings())
                 .setMessageFilter(this)
                 .build()
     }
 
-    private fun checkMessageType(messageType: String) {
-        check(messageType.isNotBlank()) { "Rule cannot be executed because the message filter does not contain 'message type'" }
+    private fun RootMessageFilter.checkRootMessageFilter() {
```
OptimumOpium, line 148: I would rather call the method `validateRootMessageFilter`

OptimumOpium (approved):

OptimumOpium (approved):
Please, use squash and merge instead of simple merge

PR #49, opened 2021-09-03
[TH2-1690] Added metrics for actual cache and rules usage

yarchiksmith:


src/main/kotlin/com/exactpro/th2/check1/metrics/BufferMetric.kt
```diff
@@ -0,0 +1,123 @@
...
+import java.util.Queue
+import java.util.LinkedList
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.ConcurrentLinkedQueue
+import java.util.concurrent.TimeUnit
+
+object BufferMetric : AutoCloseable {
+    private val logger = KotlinLogging.logger {}
+
+    private val averageReceivedMessagesSizeMetric: Gauge = Gauge
```
OptimumOpium, line 38: I don't think we need this metric. We can use the `increase` function in the Grafana to calculate the average size using the `th2_check1_actual_cache_size` and `th2_check1_actual_cache_number` metrics

src/main/kotlin/com/exactpro/th2/check1/metrics/BufferMetric.kt
```diff
@@ -0,0 +1,123 @@
...
+            .build("th2_check1_actual_cache_size", "The actual size of the cache")
+            .labelNames(DEFAULT_SESSION_ALIAS_LABEL_NAME, DEFAULT_DIRECTION_LABEL_NAME)
+            .register()
+    private val requiredCacheMemorySizeMetric: Gauge = Gauge
+            .build("th2_check1_required_cache_memory_size", "Required cache memory size without active cache size")
+            .register()
+
+    private lateinit var lastReceivedMessageMetricTask: Disposable
+    private val lastReceivedMessages: Queue<Long> = ConcurrentLinkedQueue<Long>()
+    private val bufferMessagesSizeBySessionKey: MutableMap<SessionKey, LinkedList<Long>> = ConcurrentHashMap()
```
OptimumOpium, line 55: You don't need a LinkedList here. You need a Queue

src/main/kotlin/com/exactpro/th2/check1/metrics/BufferMetric.kt
```diff
@@ -0,0 +1,123 @@
...
+            Observable.interval(messageSizeCollectionTimeout, TimeUnit.MINUTES)
+                    .subscribe {
+                        averageReceivedMessagesSizeMetric.set(lastReceivedMessages.average())
+                        lastReceivedMessages.clear()
+                    }
+
+    private fun calculateRequiredMemory(): Double {
+        return bufferMessagesSizeBySessionKey.values.sumByDouble {
+            val averageMessageSize = it.average()
+            val leftBufferCapacity = maxBufferSize - it.size
```
OptimumOpium, line 109: What do we compute here? How much memory will be additionally consumed when we fill up the buffer?

Maybe we should calculate how much memory needs the buffer when it will be filled up (averageMessageSize * maxBufferSize)? 

Also, I think we can compute that using Grafana's functions

src/main/kotlin/com/exactpro/th2/check1/metrics/RuleMetric.kt
```diff
@@ -0,0 +1,48 @@
...
+    private val logger = KotlinLogging.logger {}
+
+    private val ACTIVE_TASK_COUNTER = Gauge.build("th2_check1_active_tasks_count", "Current active tasks count in execution").register()
+    private val ACTIVE_TASK_MEMORY_SIZE = Gauge.build("th2_check1_active_tasks_memory_usage_size", "Memory usage of currently executing tasks").register()
+
+    private val activeRules: MutableMap<AbstractCheckTask, Long> = ConcurrentHashMap()
+
+    fun incrementActiveRule(rule: AbstractCheckTask) {
+        ACTIVE_TASK_COUNTER.inc()
+        val parsedInstance = GraphLayout.parseInstance(rule)
```
OptimumOpium, line 34: The size of the rule at the beginning might not match the size at the end. I am not sure if we need that metric here...

src/main/kotlin/com/exactpro/th2/check1/metrics/RuleMetric.kt
```diff
@@ -0,0 +1,48 @@
...
+        }
+        parsedInstance.totalSize().also {
+            activeRules[rule] = it
+            ACTIVE_TASK_MEMORY_SIZE.inc(it)
+        }
+    }
+
+    fun decrementActiveRule(rule: AbstractCheckTask) {
+        ACTIVE_TASK_COUNTER.dec()
+        activeRules.remove(rule)?.also { ACTIVE_TASK_MEMORY_SIZE.dec(it) }
```
OptimumOpium, line 46: The rule will be actually removed only the cleanup method in CollectorService will be called. So until then, the rule will be kept in memory

src/main/kotlin/com/exactpro/th2/check1/utils/MetricUtils.kt
```diff
@@ -0,0 +1,23 @@
+/*
```
OptimumOpium, line 1: I would move that file to `com.exactpro.th2.check1.metrics.utils`

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -486,5 +489,27 @@ abstract class AbstractCheckTask(
...
+        return true
+    }
+
+    override fun hashCode(): Int {
+        var result = description?.hashCode() ?: 0
+        result = 31 * result + timeout.hashCode()
+        result = 31 * result + sessionKey.hashCode()
+        result = 31 * result + parentEventID.hashCode()
+        return result
+    }
```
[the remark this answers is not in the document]
yarchiksmith: Then why we should use hashCode here https://github.com/th2-net/th2-check1/blob/master/src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt#L157 ?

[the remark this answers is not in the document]
OptimumOpium: We use `hashCode` at line 157 to identify the current rule that has produces that line in the log. If we don't override the `hashCode` method it will return a different value for every new instance (even if they have the same timeout, session key, etc). Each rule is unique and the rule should be equal only to itself

README.md
```diff
@@ -113,7 +112,8 @@ spec:
 
 ### 3.8.0
 
-+ Added memory volume metric for cache monitoring
++ Added memory volume metric for cache monitoring such as actual buffer count and size
```
OptimumOpium, line 115: I think we should add the actual names for metrics and describe them in the README file. Otherwise, we will have to go to the code to get the metrics' names.
Also, it would be nice to have the description for labels that are used in these metrics

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -378,6 +381,7 @@ abstract class AbstractCheckTask(
     companion object {
         const val DEFAULT_SEQUENCE = Long.MIN_VALUE
         private val RESPONSE_EXECUTOR = ForkJoinPool.commonPool()
+        private val ACTIVE_TASK_COUNTER = Gauge.build("th2_check1_active_tasks_count", "Current active tasks count in execution").register()
```
OptimumOpium, line 384: I think we should extract it into a separate object as it was before

src/main/kotlin/com/exactpro/th2/check1/metrics/BufferMetric.kt
```diff
@@ -0,0 +1,88 @@
...
+import org.openjdk.jol.info.GraphLayout
+import java.util.Queue
+import java.util.LinkedList
+import java.util.concurrent.ConcurrentHashMap
+
+object BufferMetric {
+    private val logger = KotlinLogging.logger {}
+
+    private val actualBufferCountMetric: Gauge = Gauge
+            .build("th2_check1_actual_cache_number", "The actual number of caches")
```
OptimumOpium, line 35:
```suggestion
            .build("th2_check1_actual_cache_number", "The actual number of messages in caches")
```

src/main/kotlin/com/exactpro/th2/check1/metrics/BufferMetric.kt
```diff
@@ -0,0 +1,88 @@
...
+            .labelNames(DEFAULT_SESSION_ALIAS_LABEL_NAME, DEFAULT_DIRECTION_LABEL_NAME)
+            .register()
+    private val actualBufferSizeMetric: Gauge = Gauge
+            .build("th2_check1_actual_cache_size", "The actual size of the cache")
```
OptimumOpium, line 39:
```suggestion
            .build("th2_check1_actual_cache_size", "The actual size of messages in caches")
```

src/main/kotlin/com/exactpro/th2/check1/metrics/BufferMetric.kt
```diff
@@ -0,0 +1,88 @@
...
+
+            bufferedMessageSize.add(calculatedMessageSize)
+            incrementStats(calculatedMessageSize, labels)
+
+            return@compute bufferedMessageSize
+        }
+    }
+
+    private fun calculateMessageSize(message: Message): Long {
+        val parsedInstance = GraphLayout.parseInstance(message)
```
OptimumOpium, line 72: Please, check the performance of this method. Maybe it will slow down us in message processing

OptimumOpium (changes requested):

src/main/kotlin/com/exactpro/th2/check1/metrics/BufferMetric.kt
Nikita-Smirnov-Exactpro: Please run this method vs several messages with different sizes.
Interesting How long is one message calculated.

src/main/kotlin/com/exactpro/th2/check1/metrics/BufferMetric.kt
```diff
@@ -15,74 +15,41 @@ package com.exactpro.th2.check1.metrics
...
-        }
-    }
+            actualBufferCountMetric.labels(*labels).inc()
 
-    private fun calculateMessageSize(message: Message): Long {
-        val parsedInstance = GraphLayout.parseInstance(message)
-        if (logger.isTraceEnabled) {
-            logger.trace("Foot print for message with id: '{}'\n{}", message.metadata.id.toJson(), parsedInstance.toFootprint())
+            return@compute bufferedMessagesNumber.inc()
         }
```
OptimumOpium: Looks to me that we can replace all this with the this function. It's a bit clearer in my opinion.
```kotlin
bufferMessagesSizeBySessionKey.compute(sessionKey) { _, old ->
  min(maxBufferSize, (old ?: 0) + 1).also {
    if (it != old) {
       actualBufferCountMetric.labels(*labels).inc()
    }
  }
}
```

OptimumOpium (dismissed):

build.gradle
```diff
@@ -177,6 +177,9 @@ dependencies {
 
     implementation "io.reactivex.rxjava2:rxjava:2.2.19" // https://github.com/salesforce/reactive-grpc/issues/202
 
+    // metrics
+    implementation 'io.prometheus:simpleclient'
```
OptimumOpium, line 181:
```suggestion
    implementation('io.prometheus:simpleclient') {
      because('metrics from messages and rules')
    }
```

src/main/kotlin/com/exactpro/th2/check1/CollectorService.kt
```diff
@@ -62,12 +63,18 @@ class CollectorService(
...
         val limitSize = configuration.messageCacheSize
         mqSubject = PublishSubject.create()
 
         subscriberMonitor = subscribe(MessageListener { _: String, batch: MessageBatch -> mqSubject.onNext(batch) })
         streamObservable = mqSubject.flatMapIterable(MessageBatch::getMessagesList)
-            .groupBy { message -> message.metadata.id.run { SessionKey(connectionId.sessionAlias, direction) } }
+                .groupBy { message ->
+                    message.metadata.id.run {
+                        SessionKey(connectionId.sessionAlias, direction)
+                    }.also { BufferMetric.processMessage(it) }
```
OptimumOpium, line 76:
```suggestion
                    }.also(BufferMetric::processMessage)
```

src/main/kotlin/com/exactpro/th2/check1/metrics/BufferMetric.kt
```diff
@@ -0,0 +1,49 @@
...
+import com.exactpro.th2.check1.configuration.Check1Configuration
+import com.exactpro.th2.common.metrics.DEFAULT_DIRECTION_LABEL_NAME
+import com.exactpro.th2.common.metrics.DEFAULT_SESSION_ALIAS_LABEL_NAME
+import io.prometheus.client.Gauge
+import java.util.concurrent.ConcurrentHashMap
+import kotlin.math.min
+
+object BufferMetric {
+
+    private val actualBufferCountMetric: Gauge = Gauge
```
OptimumOpium, line 26: I think we can replace Gauge with Counter because our metric does not go down (at least for now). We only increment the value

src/main/kotlin/com/exactpro/th2/check1/metrics/utils/MetricUtils.kt
```diff
@@ -0,0 +1,23 @@
+/*
```
OptimumOpium, line 1: Do we still need this class?

src/main/kotlin/com/exactpro/th2/check1/metrics/RuleMetric.kt
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.exactpro.th2.check1.metrics
+
+import io.prometheus.client.Gauge
+
+object RuleMetric {
+    private val ACTIVE_TASK_COUNTER = Gauge.build("th2_check1_active_tasks_number", "Current active tasks number in execution").register()
```
OptimumOpium, line 19: Let's add a bit of clarification to the help message. `execution` says nothing to me and it make the description hard to understand.
```suggestion
    private val ACTIVE_TASK_COUNTER = Gauge.build("th2_check1_active_tasks_number", "The number of currently working rules").register()
```
Or you can suggest another option to the description. The main point - it should help to understand what does the metric's value means.

Also, I was thinking about adding the label with rule type (CheckSequenceRule, CheckRule, etc). But maybe it is redundant for now

yarchiksmith: I think it is a good idea to control which rules are working right now

OptimumOpium (approved):

PR #50, opened 2021-09-06
[TH2-2298] Added exception handling for regex operation in the message filter

yarchiksmith:


src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -459,6 +472,22 @@ abstract class AbstractCheckTask(
         }
     }
 
+    protected fun ProtoToIMessageConverter.fromProtoPreFilter(protoPreMessageFilter: RootMessageFilter,
+                                                              messageName: String = protoPreMessageFilter.messageType): IMessage {
+        try {
+            return fromProtoFilter(protoPreMessageFilter.messageFilter, messageName)
+        } catch (e: PatternSyntaxException) {
```
OptimumOpium, line 479: There might be other exceptions (e.g. NuberFormatException). So, I think we should catch any exception that is thrown during the filter creation

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -99,6 +106,7 @@ abstract class AbstractCheckTask(
         BEGIN,
         TIMEOUT,
         COMPLETED,
+        INTERNAL_ERROR,
```
OptimumOpium, line 109: I think we should not add any new states for handling the internal error. If we got an error during initialization we should complete the rule immediately. The initialization happens in the `begin` method. So, I think we can catch the exception that is thrown during calling `.subscribe(this)` at line 255. If any exception is caught we should add information about the cause to the event, send the event and stop the rule execution

OptimumOpium (changes requested):

src/main/kotlin/com/exactpro/th2/check1/rule/RuleFactory.kt
```diff
@@ -0,0 +1,198 @@
...
+                            request.preFilter,
+                            protoMessageFilters,
+                            request.checkOrder,
+                            request.parentEventId,
+                            streamObservable,
+                            eventBatchRouter
+                    )
+                }
+                onErrorEvent(request.parentEventId) {
+                    Event.from(Instant.now())
```
Nikita-Smirnov-Exactpro, line 116:
```suggestion
                    Event.start()
```

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -219,31 +228,49 @@ abstract class AbstractCheckTask(
...
+                    .mapToMessageContainer()
+                    .taskPipeline()
+                    .subscribe(this)
+        } catch (exception: Exception) {
+            val initialException = getInitialException(exception)
+            rootEvent.addSubEventWithSamePeriod()
+                    .name("An error occurred while executing rule")
+                    .type("internalError")
+                    .status(FAILED)
+                    .bodyData(EventUtils.createMessageBean(initialException.message))
```
Nikita-Smirnov-Exactpro, line 261: Please attach the source exception via the Event.exception(exception, true) call

src/main/kotlin/com/exactpro/th2/check1/exception/RuleCreationException.kt
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+package com.exactpro.th2.check1.exception
+
+class RuleCreationException : RuleException {
+    constructor() : super()
+    constructor(message: String) : super(message)
+    constructor(message: String, cause: Throwable) : super(message, cause)
+    constructor(cause: Throwable) : super(cause)
+}
```
OptimumOpium, line 21:
```suggestion
class RuleCreationException @JvmOverloads construtor(message: String? = null, cause: Throwable? = null) : RuleException(message, cause)
```

src/main/kotlin/com/exactpro/th2/check1/exception/RuleException.kt
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.exactpro.th2.check1.exception
+
+open class RuleException : Exception {
```
OptimumOpium, line 16: I think you can move all exceptions in a single kotlin file

src/main/kotlin/com/exactpro/th2/check1/rule/RuleFactory.kt
```diff
@@ -0,0 +1,198 @@
...
+import io.reactivex.Observable
+import mu.KotlinLogging
+import org.slf4j.Logger
+import java.time.Instant
+import java.util.concurrent.ForkJoinPool
+
+class RuleFactory(
+        private val maxEventBatchContentSize: Int,
+        private val streamObservable: Observable<StreamContainer>,
+        private val eventBatchRouter: MessageRouter<EventBatch>
```
OptimumOpium, line 45: According to kotin convention, we should use the same size for indentation in the continuation indent and the normal indent. So, it would nice if you reformat the code according that
```suggestion
    private val maxEventBatchContentSize: Int,
    private val streamObservable: Observable<StreamContainer>,
    private val eventBatchRouter: MessageRouter<EventBatch>
```

src/main/kotlin/com/exactpro/th2/check1/rule/RuleFactory.kt
```diff
@@ -0,0 +1,198 @@
...
+                }
+                onErrorEvent(request.parentEventId) {
+                    Event.start()
+                            .name("Sequence check rule cannot be created")
+                            .type("sequenceCheckRuleCreation")
+                }
+            }
+
+
+    private fun <T : GeneratedMessageV3, R : AbstractCheckTask> ruleCreation(request: T, block: RuleCreationContext<T, R>.() -> Unit): R {
```
OptimumOpium, line 123: You can make this function `inline`

src/main/kotlin/com/exactpro/th2/check1/rule/RuleFactory.kt
```diff
@@ -0,0 +1,198 @@
...
+                            maxEventBatchContentSize,
+                            request.preFilter,
+                            protoMessageFilters,
+                            request.checkOrder,
+                            request.parentEventId,
+                            streamObservable,
+                            eventBatchRouter
+                    )
+                }
+                onErrorEvent(request.parentEventId) {
```
OptimumOpium, line 115: What is the point of passing `request.parentEventId` in this method? You don't use other values. You don't use that ID in this method (only store it into a field). Maybe we should pass that value into `ruleCreation` mathod?

src/main/kotlin/com/exactpro/th2/check1/rule/RuleFactory.kt
```diff
@@ -0,0 +1,198 @@
...
+
+    private fun <T : GeneratedMessageV3, R : AbstractCheckTask> ruleCreation(request: T, block: RuleCreationContext<T, R>.() -> Unit): R {
+        val ruleCreationContext = RuleCreationContext<T, R>().apply(block)
+        try {
+            return ruleCreationContext.action(request)
+        } catch (e: RuleInternalException) {
+            throw e
+        } catch (e: Exception) {
+            val rootEvent = ruleCreationContext.event()
+            rootEvent.addSubEventWithSamePeriod()
```
OptimumOpium, line 131: Please, add the exception to the event. There is a method in Event class that allows you to do that

src/main/kotlin/com/exactpro/th2/check1/rule/RuleFactory.kt
```diff
@@ -0,0 +1,198 @@
...
+                LOGGER.error("Can not send event batch '{}'", batch.toJson(), e)
+            }
+        }
+    }
+
+    companion object {
+        @JvmField
+        val LOGGER: Logger = KotlinLogging.logger { }
+        private val RESPONSE_EXECUTOR = ForkJoinPool.commonPool()
+    }
```
OptimumOpium, line 182: The companion object should be the last one in class

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -76,8 +79,13 @@ abstract class AbstractCheckTask(
...
     protected val converter = ProtoToIMessageConverter(VerificationUtil.FACTORY_PROXY, null, null)
-    protected val rootEvent: Event = Event.from(submitTime)
-        .description(description)
+    protected val rootEvent: Event by lazy {
+        Event.from(submitTime)
+                .name(name())
+                .type(type())
+                .bodyData(EventUtils.createMessageBean(eventBody()))
+                .description(description)
+    }
```
OptimumOpium, line 88: Why do we need this event to be lazy?

yarchiksmith: If we do just a normal initialization, then we will receive warnings `Calling non-final function name in constructor` in the methods `name()`, `type()`, `eventBody()`. This is the way to avoid it

OptimumOpium: Okay, I get it. But still, it is a hack to avoid warnings. If we will need to get access to the `rootEvent` in the constructor later those methods will be invoked anyway.
I think we can create a `setup` function that received the `Event` as a parameter and the rule implementation can configure the event (we can invoke it here instead of adding abstract method `eventBody()`). The methods for name and type might be useful. But the event's body might contain additional information. So, we should not restrict the ability to configure the root event with this method

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -373,6 +368,11 @@ abstract class AbstractCheckTask(
         }
     }
 
+    private fun completeRootEvent() {
```
OptimumOpium, line 371: I think `setupRootEvent` or `configureRootEvent` would be a better name

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -337,6 +364,18 @@ abstract class AbstractCheckTask(
...
     }
 
+    private fun tryToCompleteEvent(prevState: State) {
+        try {
+            completeEvent(prevState == State.TIMEOUT)
+        } catch (e: Exception) {
+            rootEvent.addSubEventWithSamePeriod()
+                    .name("Event cannot be completed")
+                    .type("eventNotComplete")
+                    .bodyData(EventUtils.createMessageBean(e.message))
```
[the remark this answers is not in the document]
OptimumOpium: There is a method in Event class to add the exception with all causes

src/main/kotlin/com/exactpro/th2/check1/rule/RuleFactory.kt
```diff
@@ -0,0 +1,195 @@
...
+        }
+
+        fun onErrorEvent(block: () -> Event) {
+            event = block
+        }
+    }
+
+    companion object {
+        @JvmField
+        val LOGGER: Logger = KotlinLogging.logger { }
```
OptimumOpium, line 192: Can be private

OptimumOpium:
Please, resolve the conflict first

OptimumOpium (dismissed):

OptimumOpium (approved):

PR #72, opened 2021-09-30
[TH2-1612] Provided more detailed logging in comparable messages

yarchiksmith:


src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -391,21 +392,26 @@ abstract class AbstractCheckTask(
...
             }
             return AggregatedFilterResult.EMPTY
         }
         val comparisonResult: ComparisonResult? = messageFilter.let {
             MessageComparator.compare(messageContainer.sailfishMessage, it.message, it.comparatorSettings, matchNames)
         }
-        LOGGER.debug("Compare message '{}' result\n{}", messageContainer.sailfishMessage.name, comparisonResult)
+
+        if (comparisonResult == null) {
+            LOGGER.debug("Comparison result for message '{}' does not match the filter by key fields or message type", messageContainer.sailfishMessage.name)
```
OptimumOpium, line 411: I would also add the expected message type from the filter. It will show if the result is null because types are different

OptimumOpium (approved):

PR #74, opened 2021-10-01
[TH2-2394] Added the ability to verify repeating groups according to defined filters

yarchiksmith:


src/test/kotlin/com/exactpro/th2/check1/rule/check/TestCheckRuleTask.kt
```diff
@@ -348,6 +349,97 @@ internal class TestCheckRuleTask : AbstractCheckTaskTest() {
...
+
+            expectedLegs.forEach { (leg, verificationEntryByField) ->
+                val actualLeg = actualLegs[leg].fields
+                assertNotNull(actualLeg) { "The validation event does not contain the expected leg" }
+                verificationEntryByField.forEach { (field, status) ->
+                    val expectedVerificationEntry = actualLeg[field]
+                    assertNotNull(expectedVerificationEntry) { "Actual leg does not contain the expected field" }
+                    assertEquals(status, expectedVerificationEntry.status)
+                }
+            }
```
OptimumOpium, line 439: I have already so code like this in one of the tests. Maybe we should extract the logic into a separate method?

src/test/kotlin/com/exactpro/th2/check1/rule/check/TestCheckRuleTask.kt
```diff
@@ -407,40 +400,57 @@ internal class TestCheckRuleTask : AbstractCheckTaskTest() {
...
+            assertVerificationEntriesLegsByStatus(expectedLegs, actualLegs)
         })
     }
 
 
+    private fun extractVerificationFromEvent(verificationEvent: Event): Verification? {
+        return jacksonObjectMapper().readValue<List<Verification>>(verificationEvent.body.toByteArray()).firstOrNull()
+    }
+
+    private fun assertVerificationsLegs(verificationEvent: Event): List<VerificationEntry> {
```
OptimumOpium, line 433: I think we should return the whole verification to check it as we need.

src/test/kotlin/com/exactpro/th2/check1/rule/check/TestCheckRuleTask.kt
```diff
@@ -407,40 +400,57 @@ internal class TestCheckRuleTask : AbstractCheckTaskTest() {
...
+        val verification = extractVerificationFromEvent(verificationEvent)
+        assertNotNull(verification) { "Verification event does not contain the verification" }
+        val actualLegs = verification.fields["legs"]?.fields?.values?.toList()
+        assertNotNull(actualLegs) { "Actual legs is missed" }
+        return actualLegs
+    }
+
+    private fun assertVerificationEntriesLegsByStatus(expectedLegs: List<VerificationEntry>, actualLegs: List<VerificationEntry>) {
```
OptimumOpium, line 441: I think it will be better to compare whole verifications here instead of VerificationEntry's

src/test/kotlin/com/exactpro/th2/check1/rule/check/TestCheckRuleTask.kt
```diff
@@ -407,40 +400,57 @@ internal class TestCheckRuleTask : AbstractCheckTaskTest() {
...
-                    assertEquals(status, expectedVerificationEntry.status)
-                }
-            }
+            assertVerificationEntriesLegsByStatus(expectedLegs, actualLegs)
         })
     }
 
 
+    private fun extractVerificationFromEvent(verificationEvent: Event): Verification? {
+        return jacksonObjectMapper().readValue<List<Verification>>(verificationEvent.body.toByteArray()).firstOrNull()
```
OptimumOpium, line 430: That won't work in case the description was added to the event's body. As a solution, we can use the `addMixIn` functionality in Jackson.
We will need to create an interface and annotate it with `@JsonSubTypes` and `@JsonTypeInfo` annotations. It will allow us to register the Verification and com.exactpro.th2.common.event.bean.Message classes  as possible options for `IBodyData`

src/test/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTaskTest.kt
```diff
@@ -51,6 +61,56 @@ abstract class AbstractCheckTaskTest {
...
+            asserts: (VerificationEntry, VerificationEntry) -> Unit) {
+        assertNotNull(actualVerificationEntries) { "Actual verification entry is null" }
+        expectedVerificationEntries.forEach { (expectedFieldName, expectedVerificationEntry) ->
+            val actualVerificationEntry = actualVerificationEntries[expectedFieldName]
+            assertNotNull(actualVerificationEntry) {
+                "Actual verification entry does not contains field '${expectedFieldName}'"
+            }
+            asserts(expectedVerificationEntry, actualVerificationEntry)
+            if (expectedVerificationEntry.fields == null) {
+                return
```
OptimumOpium, line 89: The `forEach` function is inline one. Calling `return` will cause exiting the whole loop instead of skipping a single iteration
```suggestion
                return@forEach
```

OptimumOpium (approved):

PR #75, opened 2021-10-01
[TH2-2437] Auto silence check after the CheckSequenceRule finds all messages

OptimumOpium:


src/main/kotlin/com/exactpro/th2/check1/CollectorService.kt
```diff
@@ -105,9 +106,17 @@ class CollectorService(
...
+
+        val silenceCheckTask: AbstractCheckTask? = if (silenceCheck) {
+            ruleFactory.createSilenceCheck(request, olderThanTimeUnit.duration.toMillis() * olderThanDelta)
+        } else {
+            null
+        }
 
         eventIdToLastCheckTask.compute(CheckTaskKey(chainID, request.connectivityId)) { _, value ->
             task.apply { addToChainOrBegin(value, request.checkpoint) }
+                .run { silenceCheckTask?.also { subscribeNextTask(it) } ?: this }
```
Nikita-Smirnov-Exactpro, line 119:
```suggestion
                .run {
                    val silenceCheck = if (request.hasSilenceCheck()) request.silenceCheck.value else defaultAutoSilenceCheck
                    if (silenceCheck) {
                        val silenceCheckTask = ruleFactory.createSilenceCheck(request, olderThanTimeUnit.duration.toMillis() * olderThanDelta)
                        subscribeNextTask(silenceCheckTask)
                        return@run silenceCheckTask
                    }
                    return@run this
                }
```
Maybe refactor like above

src/main/kotlin/com/exactpro/th2/check1/rule/sequence/SilenceCheckTask.kt
```diff
@@ -0,0 +1,152 @@
...
+
+        rootEvent.addSubEvent(preFilterEvent)
+
+        resultEvent = Event.start()
+            .type("noMessagesCheckResult")
+        rootEvent.addSubEvent(resultEvent)
+    }
+
+    override fun onChainedTaskSubscription() {
+        if (started) {
```
Nikita-Smirnov-Exactpro, line 101: This check doesn't guarantee that the start method is called at this moment and will be continued after this check.
I think we should call cancel and checkComplete methods always and start method shouldn't be changed

src/main/kotlin/com/exactpro/th2/check1/rule/sequence/SilenceCheckTask.kt
```diff
@@ -0,0 +1,152 @@
...
+            converter.fromMetadataFilter(it, VerificationUtil.METADATA_MESSAGE_NAME),
+            it.toComparisonSettings()
+        )
+    }
+    private lateinit var preFilterEvent: Event
+    private lateinit var resultEvent: Event
+    private var extraMessagesCounter: Int = 0
+
+    @Volatile
+    private var started = false
```
Nikita-Smirnov-Exactpro, line 73: Redundant because it guarantee nothing

Nikita-Smirnov-Exactpro (changes requested):

build.gradle
```diff
@@ -165,7 +166,7 @@ signing {
 
 dependencies {
     api platform('com.exactpro.th2:bom:3.0.0')
-    implementation 'com.exactpro.th2:grpc-check1:3.3.0'
+    implementation 'com.exactpro.th2:grpc-check1:3.5.0-th2-2437-1292014101-SNAPSHOT'
```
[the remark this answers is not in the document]
OptimumOpium: Yes, I remember. It is a draft PR

src/main/kotlin/com/exactpro/th2/check1/CollectorService.kt
OptimumOpium: I thought about this but the `compute` method should be as simple as possible. That's why I have decided to extract rule creation outside the `compute` method

src/main/kotlin/com/exactpro/th2/check1/rule/sequence/SilenceCheckTask.kt
```diff
@@ -148,5 +151,14 @@ class SilenceCheckTask(
             resultEvent.status(Event.Status.FAILED)
                 .name("Check failed: $extraMessagesCounter extra messages were found.")
         }
+        return true
```
Nikita-Smirnov-Exactpro, line 154: Maybe use the same approach as other rules
```suggestion
        return super.completeEvent(taskState)
```

Nikita-Smirnov-Exactpro (changes requested):

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -250,11 +267,18 @@ abstract class AbstractCheckTask(
...
      * @throws IllegalStateException when method is called more than once.
      */
-    private fun begin(sequence: Long = DEFAULT_SEQUENCE, checkpointTimestamp: Timestamp? = null,
-                      executorService: ExecutorService = createExecutorService(), untrusted: Boolean = false) {
+    private fun begin(
+        sequence: Long = DEFAULT_SEQUENCE,
+        checkpointTimestamp: Timestamp? = null,
+        executorService: ExecutorService = createExecutorService(),
+        untrusted: Boolean = false,
+        parentTaskCompleted: Boolean? = null
```
yarchiksmith, line 278: What about a new data class? Something like `ExecutionState`

src/main/kotlin/com/exactpro/th2/check1/rule/RuleFactory.kt
```diff
@@ -157,11 +159,39 @@ class RuleFactory(
...
+                    sessionKey,
+                    request.parentEventId,
+                    streamObservable,
+                    eventBatchRouter
+                )
+            }
+            onErrorEvent {
+                Event.start()
+                    .name("Auto silence check rule cannot be created")
+                    .type("checkRuleCreation")
```
yarchiksmith, line 186: I guess it should be a different type

OptimumOpium: The type groups the events that care the similar information. In our case, the information that rule was not created. So, I think we should leave the type as is (it is the same for all other rules)

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
OptimumOpium: Do you mean replacing the `untrusted` and `parentTaskCompleted` parameters with that data class?

Nikita-Smirnov-Exactpro (approved):

PR #78, opened 2021-10-06
Release 3.8.0

OptimumOpium:


src/main/java/com/exactpro/th2/check1/event/CheckSequenceUtils.java
```diff
@@ -142,6 +145,9 @@ private static String getKeyFields(String name, ValueFilter valueFilter) {
 
     private static String getKeyFields(String name, SimpleFilter valueFilter) {
         if (valueFilter.getKey()) {
+            if (valueFilter.hasSimpleList() && (valueFilter.getOperation() == FilterOperation.IN || valueFilter.getOperation() == FilterOperation.NOT_IN)) {
+                return ", " + name + ' ' + valueFilter.getOperation() + ' ' + valueFilter.getSimpleList().getSimpleValuesList();
```
Nikita-Smirnov-Exactpro, line 149: Maybe extract to a separate method and reuse

src/main/kotlin/com/exactpro/th2/check1/metrics/BufferMetric.kt
```diff
@@ -0,0 +1,49 @@
...
+        this.maxBufferSize = configuration.messageCacheSize
+    }
+
+    fun processMessage(sessionKey: SessionKey) {
+        val labels = arrayOf(sessionKey.sessionAlias, sessionKey.direction.name)
+
+        bufferMessagesSizeBySessionKey.compute(sessionKey) { _, old ->
+            min(maxBufferSize, (old ?: 0) + 1).also {
+                if (it != old) {
+                    actualBufferCountMetric.labels(*labels).inc()
```
Nikita-Smirnov-Exactpro, line 44: Maybe cache instances of the Child class to avoid the complex of actions to get a Child from metric

src/main/java/com/exactpro/th2/check1/event/CheckSequenceUtils.java
OptimumOpium: I don't see any benefits from that. In one case we operate with `ValueFilter` object and in another one with `SimpleFilter`. Those objects are different and because of that, the extracted method will have objects we need as parameters. It will make code harder to read in my opinion

Nikita-Smirnov-Exactpro: I meant the format of the result string

PR #88, opened 2021-11-10
[TH2-2654] Fixed replacement of the expected result with the string "null" value

yarchiksmith:


README.md
```diff
@@ -161,6 +161,7 @@ The `th2_check1_active_tasks_number` metric separate rules with label `rule_type
 + Migrated `grpc-check1` version from `3.4.2` to `3.5.0`
 + Migrated `sailfish-utils` version from `3.9.1` to `3.10.1`
   + Fixed conversion of `null` values
++ Now the strings with value `null` in the verification entry are converted as `null`
```
OptimumOpium, line 164:
```suggestion
+ Corrected verification entry when the `null` value and string `"null"` looked the same for the expected value
```

OptimumOpium (approved):

PR #100, opened 2021-11-24
[TH2-2458] Improvement checkpoint validation

yarchiksmith:


src/main/kotlin/com/exactpro/th2/check1/CollectorService.kt
```diff
@@ -84,7 +84,7 @@ class CollectorService(
 
         checkpointSubscriber = streamObservable.subscribeWith(CheckpointSubscriber())
 
-        ruleFactory = RuleFactory(configuration, streamObservable, eventBatchRouter)
+        ruleFactory = RuleFactory(configuration, streamObservable, eventBatchRouter, eventIdToLastCheckTask.keys)
```
OptimumOpium, line 87: There might be a case when we create a rule using factory the chain exists but when we try to add the rule into the chain it is already cleaned and removed.

So, we must get the information about the chain existing only when we try to add the rule to a chain.
Maybe we should call the factory method inside `the compute` and pass the information about the chain existing as a parameter to the method for rule creation

src/main/kotlin/com/exactpro/th2/check1/CollectorService.kt
```diff
@@ -90,11 +90,11 @@ class CollectorService(
     @Throws(InterruptedException::class)
     fun verifyCheckRule(request: CheckRuleRequest): ChainID {
         val chainID = request.getChainIdOrGenerate()
-        val task = ruleFactory.createCheckRule(request)
 
         cleanupTasksOlderThan(olderThanDelta, olderThanTimeUnit)
 
-        eventIdToLastCheckTask.compute(CheckTaskKey(chainID, request.connectivityId)) { _, value ->
+        eventIdToLastCheckTask.compute(CheckTaskKey(chainID, request.connectivityId)) { key, value ->
+            val task = ruleFactory.createCheckRule(request, eventIdToLastCheckTask.containsKey(key))
```
OptimumOpium, line 97: You already know about value for the key because you are computing one
```suggestion
            val task = ruleFactory.createCheckRule(request, value != null)
```

src/main/kotlin/com/exactpro/th2/check1/entities/RequestAdaptor.kt
```diff
@@ -0,0 +1,61 @@
...
+import com.exactpro.th2.check1.grpc.NoMessageCheckRequest
+import com.exactpro.th2.common.grpc.Checkpoint
+import com.exactpro.th2.common.grpc.ConnectionID
+
+class RequestAdaptor {
+    var hasCheckpoint: Boolean = false
+    var hasChainId: Boolean = false
+    lateinit var chainId: ChainID
+    lateinit var checkpoint: Checkpoint
+    lateinit var connectionId: ConnectionID
```
OptimumOpium, line 28: You can combine `has*` and `*` properties into one with nullable type

```kotlin
val chainId: ChainID?
```

Also, you can move these properties into constructor

src/main/kotlin/com/exactpro/th2/check1/rule/RuleFactory.kt
```diff
@@ -272,10 +272,16 @@ class RuleFactory(
         }
     }
 
-    private fun checkCheckpoint(checkpoint: Checkpoint, sessionKey: SessionKey) {
-        check(checkpoint !== Checkpoint.getDefaultInstance()) { "Request doesn't contain a checkpoint" }
+    private fun checkCheckpoint(requestAdaptor: RequestAdaptor, sessionKey: SessionKey, isChainIdExist: Boolean) {
+        if (requestAdaptor.hasChainId) {
+            check(isChainIdExist) {
+                "The request has an invalid chain id or connectivity id"
```
OptimumOpium, line 278: Please, add information that the checkpoint is required

src/main/kotlin/com/exactpro/th2/check1/rule/RuleFactory.kt
```diff
@@ -272,10 +272,16 @@ class RuleFactory(
...
+            }
+            return // We should validate checkpoint only if the request doesn't contain a chain id
+        }
+        check(requestAdaptor.hasCheckpoint) { "Request doesn't contain a checkpoint" }
```
OptimumOpium, line 282: Please, add information about why the checkpoint is required (the messageTimeout is used and no chain ID was provided)

src/test/kotlin/com/exactpro/th2/check1/rule/RuleFactoryTest.kt
```diff
@@ -69,6 +77,121 @@ class RuleFactoryTest {
...
+                .setRootFilter(RootMessageFilter.newBuilder()
+                        .setMessageType("TestMsgType")
+                )
+                .setMessageTimeout(5)
+                .setChainId(ChainID.newBuilder().setId("test_chain_id"))
+                .build()
+
+        assertThrows<RuleCreationException> {
+            ruleFactory.createCheckRule(request, false)
+        }
```
OptimumOpium, line 138: Please, add checking for exception message

src/test/kotlin/com/exactpro/th2/check1/rule/RuleFactoryTest.kt
```diff
@@ -34,8 +39,11 @@ import com.nhaarman.mockitokotlin2.verify
 import io.reactivex.Observable
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.assertAll
+import org.junit.jupiter.api.assertDoesNotThrow
 import org.junit.jupiter.api.assertThrows
+import java.time.Instant
 import kotlin.test.assertEquals
+import kotlin.test.assertNotNull
 
 class RuleFactoryTest {
```
OptimumOpium, line 48: Is there a test that checks the exception is thrown when the checkpoint does not contain the requested session alias?

yarchiksmith: Yup. The first test with the name `failed rule creation because one of required fields is empty` throws in this case

OptimumOpium: You should verify the exception message then. Could you please add it?

src/main/kotlin/com/exactpro/th2/check1/entities/RequestAdaptor.kt
```diff
@@ -20,42 +20,31 @@ import com.exactpro.th2.check1.grpc.NoMessageCheckRequest
...
-                checkpoint = request.checkpoint
-                hasChainId = request.hasChainId()
-                chainId = request.chainId
-                connectionId = request.connectivityId
-            }
+            return RequestAdaptor(
+                    if (request.hasChainId()) request.chainId else null,
+                    if (request.hasCheckpoint()) request.checkpoint else null,
+                    if (request.hasConnectivityId()) request.connectivityId else null
+            )
```
OptimumOpium, line 47: You can simplify code by using `run` method.
```kotlin
            return request.run {
                   RequestAdaptor(
                        if (hasChainId()) chainId else null,
                        if (hasCheckpoint()) checkpoint else null,
                        if (hasConnectivityId()) connectivityId else null
                     )
                }
```

src/test/kotlin/com/exactpro/th2/check1/rule/RuleFactoryTest.kt
```diff
@@ -65,10 +64,13 @@ class RuleFactoryTest {
                 .setParentEventId(EventID.newBuilder().setId("root").build())
                 .setCheckpoint(Checkpoint.newBuilder().setId(EventUtils.generateUUID()).build()).build()
 
-        assertThrows<RuleCreationException> {
+        val mainException = assertThrows<RuleCreationException> {
             ruleFactory.createCheckRule(request, true)
         }
 
+        assertEquals("An error occurred while creating rule", mainException.message)
+        assertEquals("Session alias cannot be empty", mainException.cause?.message)
```
OptimumOpium, line 72: You do the same things here and in `failed rule creation with missed checkpoint and invalid chain id` test.
You can extract this into a method and reuse part of the code

src/main/kotlin/com/exactpro/th2/check1/entities/RequestAdaptor.kt
```diff
@@ -0,0 +1,50 @@
...
+package com.exactpro.th2.check1.entities
+
+import com.exactpro.th2.check1.grpc.ChainID
+import com.exactpro.th2.check1.grpc.CheckRuleRequest
+import com.exactpro.th2.check1.grpc.CheckSequenceRuleRequest
+import com.exactpro.th2.check1.grpc.NoMessageCheckRequest
+import com.exactpro.th2.common.grpc.Checkpoint
+import com.exactpro.th2.common.grpc.ConnectionID
+
+class RequestAdaptor(val chainId: ChainID?, val checkpoint: Checkpoint?, val connectionId: ConnectionID?) {
```
OptimumOpium, line 23: It is usually better to put each property on a new line if there are more than 2 properties

src/main/kotlin/com/exactpro/th2/check1/rule/RuleFactory.kt
```diff
@@ -45,23 +45,23 @@ import java.time.Instant
 import java.util.concurrent.ForkJoinPool
 
 class RuleFactory(
-    configuration: Check1Configuration,
-    private val streamObservable: Observable<StreamContainer>,
-    private val eventBatchRouter: MessageRouter<EventBatch>
+        configuration: Check1Configuration,
+        private val streamObservable: Observable<StreamContainer>,
+        private val eventBatchRouter: MessageRouter<EventBatch>
```
OptimumOpium, line 50: According to the Kotlin code style the properties and parameters in the constructor should have a single indent (4 spaces instead of 8)

src/test/kotlin/com/exactpro/th2/check1/rule/RuleFactoryTest.kt
yarchiksmith: Sure. Also covered all cases where a rule cannot be created

OptimumOpium (approved):

PR #106, opened 2021-11-30
[TH2-2764] Add support for null markers in message

OptimumOpium:


README.md
```diff
@@ -162,6 +162,7 @@ The `th2_check1_active_tasks_number` metric separate rules with label `rule_type
 + Migrated `grpc-check1` version from `3.4.2` to `3.5.0`
 + Migrated `sailfish-utils` version from `3.9.1` to `3.10.2`
```
yarchiksmith, line 163: Update sf-utils version

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -85,7 +86,7 @@ abstract class AbstractCheckTask(
 
     protected var handledMessageCounter: Long = 0
 
-    protected val converter = ProtoToIMessageConverter(VerificationUtil.FACTORY_PROXY, null, null)
+    protected val converter = ProtoToIMessageConverter(VerificationUtil.FACTORY_PROXY, null, null, createParameters().setUseMarkerForNullsInMessage(true))
```
yarchiksmith, line 89: Can we make the `converter` as a static field?

yarchiksmith (changes requested):

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
OptimumOpium: Yep, I think there is no problem to do that.

PR #112, opened 2021-12-03
[TH2-2457] Add description for required and optional parameters in ch…

OptimumOpium:
…eck requests


README.md
```diff
@@ -24,6 +24,47 @@ Available requests are described in [this repository](https://gitlab.exactpro.co
...
+
+### CheckSequenceRuleRequest
+
+#### Required
+
+* **root_message_filters** or **message_filters** with at least one filter
+  (please note, that the `message_filters` parameter is deprecated and will be removed in the future releases)
+
+#### Optional
+* **pre_filter** - pre-filtering for messages. Only messages passed the filter will be checked by main filters.
```
jean-michel-perli-sonsin, line 59:
```suggestion
* **pre_filter** - pre-filtering for messages. Only messages that passed the filter will be checked by the main filters.
```

jean-michel-perli-sonsin (changes requested):
Hello, please see if you agree with the suggestions proposed below:
README.md
Line 21 - It verifies that there were not any messages **matching** the pre-filter in **the** original request.

Line 23 - Report about unexpected messages only after the timeout **is exceeded**. Reports nothing if any task is added to the chain.

Line 112/153 - The decimal precision is used to compare **the value of two numbers.**

Line 118/159 - MQ pin: it is used for listening to parsed messages. You can link several sources with different directions and session **aliases** to it.

Line 156/197 - New configuration parameter time-precision which **is used** if the user has not specified a time precision

Line 157/198 - New configuration parameter decimal-precision which **is used** if the user has not specified a number precision

In case of any questions, please let me know.
Thanks

jean-michel-perli-sonsin (approved):

PR #159, opened 2022-08-08
TH2-3615/3841

lumber1000:
Removing unnecessary references from rules objects that complete the work.
Failing check if check task was not started.

### DIFF
--- a/src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
+++ b/src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
@@ -81,6 +81,16 @@ abstract class AbstractCheckTask(
     private val eventBatchRouter: MessageRouter<EventBatch>
 ) : AbstractSessionObserver<MessageContainer>() {
 
+    protected class RefsKeeper<T>(refs: T) {
+        private var refsNullable: T? = refs
+        val refs: T get() = refsNullable ?: error("Requesting references after references has been erased.")
+        fun eraseRefs() {
+            refsNullable = null
+        }
+    }
+
+    protected abstract val refsKeeper: RefsKeeper<out Any>
+
     val description: String? = ruleConfiguration.description
     private val taskTimeout: TaskTimeout = ruleConfiguration.taskTimeout
 
@@ -347,7 +357,7 @@ abstract class AbstractCheckTask(
                 .build())
         } finally {
             RuleMetric.decrementActiveRule(type())
-            disposeResources()
+            refsKeeper.eraseRefs()
             sequenceSubject.onSuccess(Legacy(executorService, SequenceData(lastSequence, lastMessageTimestamp, !hasMessagesInTimeoutInterval)))
         }
     }
@@ -383,8 +393,6 @@ abstract class AbstractCheckTask(
         super.onComplete()
         end(streamCompletedState, "Message stream is completed")
     }
-
-    abstract fun disposeResources()
 
     /**
      * Prepare the root event or children events for publication.
--- a/src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
+++ b/src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
@@ -140,12 +140,20 @@ abstract class AbstractCheckTask(
     private var bufferContainsStartMessage: Boolean = false
     private var isDefaultSequence: Boolean = false
 
-    override fun onStart() {
+    @Volatile
+    protected var started = false
+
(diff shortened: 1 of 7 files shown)


src/main/kotlin/com/exactpro/th2/check1/rule/sequence/SequenceCheckRuleTask.kt
```diff
@@ -57,51 +58,64 @@ import kotlin.collections.set
...
+        val metadataPreFilter: SailfishFilter?,
+    ) {
+        val preFilterEvent: Event by lazy {
+            Event.start()
+                .type("preFiltering")
+                .bodyData(protoPreMessageFilter.toReadableBodyCollection())
+        }
 
-    private lateinit var messageFilteringResults: MutableMap<MessageID, ComparisonContainer>
+        val preFilteringResults: MutableMap<MessageID, ComparisonContainer> by lazy { LinkedHashMap() }
```
OptimumOpium, line 86: I think we can make it `val` because the `lazy` will create some objects as well. So, I don't think that using lazy for light objects is better that simple reference to the object

src/main/kotlin/com/exactpro/th2/check1/rule/sequence/SequenceCheckRuleTask.kt
```diff
@@ -57,51 +58,64 @@ import kotlin.collections.set
...
-    private lateinit var preFilterEvent: Event
+        /**
+         * List of filters which haven't matched yet. It is created from the requested filters and reduced after every match
+         */
+        lateinit var messageFilters: MutableList<MessageFilterContainer>
 
-    private var reordered: Boolean = false
-    private lateinit var matchedByKeys: MutableSet<MessageFilterContainer>
+        val messageFilteringResults: MutableMap<MessageID, ComparisonContainer> by lazy { LinkedHashMap() }
+        val matchedByKeys: MutableSet<MessageFilterContainer> by lazy { HashSet(protoMessageFilters.size) }
```
OptimumOpium, line 94: The same as for `preFilteringResults`

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -130,13 +140,21 @@ abstract class AbstractCheckTask(
     private var bufferContainsStartMessage: Boolean = false
     private var isDefaultSequence: Boolean = false
 
-    override fun onStart() {
+    @Volatile
+    protected var started = false
```
OptimumOpium, line 144:
```suggestion
    protected var started = false
      private set
```

Also, how does it solve the problem with LazyInitException? Each rule needs to check this property?

OptimumOpium (changes requested):

src/main/kotlin/com/exactpro/th2/check1/rule/check/CheckRuleTask.kt
```diff
@@ -35,53 +35,66 @@ import java.time.Instant
...
+    protected class Refs(
+        rootEvent: Event,
+        val protoMessageFilter: RootMessageFilter,
+        val messageFilter: SailfishFilter,
+        val metadataFilter: SailfishFilter?
+    ) : AbstractCheckTask.Refs(rootEvent)
+
+    override val refsKeeper = RefsKeeper(
+        Refs(
+            rootEvent = createRootEvent(),
```
OptimumOpium, line 58: Why do we create root event in each rule?

lumber1000: Because we need to pass it to Refs constructor which is called in each rule class. And we don't using inner classes (in this case we could create root event inside AbstractCheckTask.Refs class).

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -422,18 +445,27 @@ abstract class AbstractCheckTask(
...
-            doAfterCompleteEvent()
-            false
+            if (started) {
+                completeEvent(prevState)
+                doAfterCompleteEvent()
+                false
+            } else {
+                LOGGER.error("Check task was not started.")
+                refs.rootEvent.addSubEventWithSamePeriod()
+                    .name("Check failed: check task was not started.")
```
OptimumOpium, line 461: Let's add extra information to help to understand the problem
```suggestion
                    .name("Check failed: task timeout elapsed before the check task was started. Please, check component resources for throttling or intensive GC")
```

OptimumOpium (changes requested):

PR #160, opened 2022-08-19
Support for disabling of order verification for simple collection

lumber1000:


src/main/java/com/exactpro/th2/check1/configuration/Check1Configuration.java
```diff
@@ -88,4 +91,8 @@ public Duration getTimePrecision() {
     public boolean isCheckNullValueAsEmpty() {
         return checkNullValueAsEmpty;
     }
+
+    public boolean getDefaultCheckSimpleCollectionsOrder() {
```
OptimumOpium, line 95:
```suggestion
    public boolean isDefaultCheckSimpleCollectionsOrder() {
```

OptimumOpium (changes requested):
Could you please add tests for new functionality?

src/test/kotlin/com/exactpro/th2/check1/rule/RuleFactoryTest.kt
```diff
@@ -413,6 +416,67 @@ class RuleFactoryTest {
...
+            val messageFilterField = this.javaClass.getDeclaredField("messageFilter")
+            messageFilterField.isAccessible = true
+            return messageFilterField.get(this) as SailfishFilter
+        }
+
+    private fun Check1Configuration.setDefaultCheckSimpleCollectionsOrder(value: Boolean) {
+        val defaultCheckField = this.javaClass.getDeclaredField("defaultCheckSimpleCollectionsOrder")
+        defaultCheckField.isAccessible = true
+        defaultCheckField.setBoolean(this, value)
+    }
```
OptimumOpium, line 478: Why do you use accessors to private fields? If you need this then something wrong with the code and we need to change it.

Also, I think we should check the work of the created rule and see if it behaves like we expect

OptimumOpium (changes requested):

build.gradle
```diff
@@ -167,7 +167,7 @@ signing {
 dependencies {
     api platform('com.exactpro.th2:bom:3.0.0')
     implementation 'com.exactpro.th2:grpc-check1:3.5.1'
```
OptimumOpium, line 169: We need to add dependecy to a new grpc-common version in grpc-check1. And update the grpc-check1 version here

OptimumOpium (changes requested):

build.gradle
```diff
@@ -165,23 +165,23 @@ signing {
 }
 
 dependencies {
-    api platform('com.exactpro.th2:bom:3.0.0')
-    implementation 'com.exactpro.th2:grpc-check1:3.5.1'
-    implementation 'com.exactpro.th2:common:3.31.3'
-    implementation 'com.exactpro.th2:sailfish-utils:3.12.3'
+    api platform('com.exactpro.th2:bom:3.2.0')
+    implementation 'com.exactpro.th2:grpc-check1:3.6.1-TH2-3775-2910563097-SNAPSHOT'
+    implementation 'com.exactpro.th2:common:3.41.0-TH2-3775-2889888066-SNAPSHOT'
```
OptimumOpium, line 170: Please, use the last release version for common

src/test/kotlin/com/exactpro/th2/check1/rule/RuleFactoryTest.kt
OptimumOpium: Any updates about that?

build.gradle
```diff
@@ -165,23 +165,23 @@ signing {
 }
 
 dependencies {
-    api platform('com.exactpro.th2:bom:3.0.0')
-    implementation 'com.exactpro.th2:grpc-check1:3.5.1'
-    implementation 'com.exactpro.th2:common:3.31.3'
-    implementation 'com.exactpro.th2:sailfish-utils:3.12.3'
+    api platform('com.exactpro.th2:bom:3.2.0')
+    implementation 'com.exactpro.th2:grpc-check1:3.6.1-TH2-3775-2910563097-SNAPSHOT'
```
OptimumOpium, line 169: Please, update to dev version

OptimumOpium (changes requested):

src/test/kotlin/com/exactpro/th2/check1/rule/RuleFactoryTest.kt
OptimumOpium: Please, correct tests. We should not use any reflection and access to private fields in tests

OptimumOpium (changes requested):

src/test/kotlin/com/exactpro/th2/check1/rule/check/TestCheckRuleTask.kt
```diff
@@ -307,6 +314,85 @@ internal class TestCheckRuleTask : AbstractCheckTaskTest() {
         })
     }
 
+    @ParameterizedTest
+    @ValueSource(ints = [0, 1, 2, 3, 4])
+    fun `simple collection order checking config`(testCase: Int) {
```
OptimumOpium, line 319: Instead of this you can use `@MethodSource` annotation and provide the required arguments in a method instead of using an array inside the method

OptimumOpium (changes requested):

OptimumOpium (approved):

PR #184, opened 2023-02-02
[TH2-4728] Method for waiting for the rule result

lumber1000:
WaitForResult method implementation


.github/workflows/dev-docker-publish.yml
```diff
@@ -35,8 +35,8 @@ jobs:
       - uses: docker/login-action@v1
         with:
           registry: ghcr.io
-          username: ${{ github.repository_owner }}
-          password: ${{ secrets.CR_PAT }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
```
OptimumOpium, line 39: Please, sink with master instead making those changes to workflows

src/main/java/com/exactpro/th2/check1/Check1Handler.java
```diff
@@ -68,16 +77,18 @@ public void createCheckpoint(CheckpointRequest request, StreamObserver<Checkpoin
...
                 logger.info("Submit CheckRule request: " + MessageUtils.toJson(request));
             }
 
             CheckRuleResponse.Builder response = CheckRuleResponse.newBuilder();
             try {
-                ChainID chainID = collectorService.verifyCheckRule(request);
+                ChainID chainID = collectorService.verifyCheckRule(request, ruleId);
                 response.setChainId(chainID)
                         .setStatus(RequestStatus.newBuilder().setStatus(SUCCESS));
+                if (ruleId != 0) response.setRuleId(ruleId);
```
OptimumOpium, line 91: It won't make any difference because the ID is an int in grpc message. So, it is 0 by default

src/main/java/com/exactpro/th2/check1/Check1Handler.java
```diff
@@ -167,11 +184,44 @@ public void submitNoMessageCheck(NoMessageCheckRequest request, StreamObserver<N
...
+                    responseBuilder.setRuleResult(ruleResult);
+                    requestStatusBuilder.setStatus(SUCCESS);
+                } else {
+                    requestStatusBuilder.setStatus(ERROR).setMessage("No rule with specified id found");
+                }
+            } catch (TimeoutException e) {
+                if (logger.isDebugEnabled()) {
+                    logger.debug("WaitForResult timeout expired");
+                }
+                requestStatusBuilder.setStatus(ERROR).setMessage("Timeout expired");
```
OptimumOpium, line 214: I think we should encapsulate that in the storage itself. We can create sealed class `RuleResult` and several implementations `Executed(val status: Status)`, `Timeout`, `NotFound` and use them here to create a response

src/main/java/com/exactpro/th2/check1/Check1Handler.java
```diff
@@ -167,11 +184,44 @@ public void submitNoMessageCheck(NoMessageCheckRequest request, StreamObserver<N
...
+                if (ruleResult != null) {
+                    responseBuilder.setRuleResult(ruleResult);
+                    requestStatusBuilder.setStatus(SUCCESS);
+                } else {
+                    requestStatusBuilder.setStatus(ERROR).setMessage("No rule with specified id found");
+                }
+            } catch (TimeoutException e) {
+                if (logger.isDebugEnabled()) {
+                    logger.debug("WaitForResult timeout expired");
+                }
```
OptimumOpium, line 213: No needs to wrap logging into `if` statement if we use constant string or use logger formatting. The formatting will be evaluated only if logging level is enabled

src/main/java/com/exactpro/th2/check1/Check1Main.java
```diff
@@ -42,10 +43,11 @@ public static void main(String[] args) {
             GrpcRouter grpcRouter = commonFactory.getGrpcRouter();
             Check1Configuration configuration = commonFactory.getCustomConfiguration(Check1Configuration.class);
 
-            CollectorService collectorService = new CollectorService(messageRouter, commonFactory.getEventBatchRouter(), configuration);
+            ResultsStorage resultsStorage = new ResultsStorage(configuration.getResultsCleanupTimeoutSec());
```
OptimumOpium, line 46: I think we should use the same parameter that we use for rule clean up. Don't see any reason to add additional parameter for that

src/main/kotlin/com/exactpro/th2/check1/ResultsStorage.kt
```diff
@@ -0,0 +1,52 @@
...
+import com.exactpro.th2.common.grpc.EventStatus
+import java.util.concurrent.CompletableFuture
+import java.util.concurrent.Executors
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.TimeoutException
+import java.util.concurrent.atomic.AtomicLong
+
+class ResultsStorage(private val cleanupTimeoutSec: Long) : AutoCloseable {
+    private val idCounter = AtomicLong()
+    private val results: MutableMap<Long, CompletableFuture<EventStatus>> = HashMap()
```
OptimumOpium, line 28: Probably, it should be a ConcurrentHashMap. We can access it from different threads

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -78,10 +83,11 @@ abstract class AbstractCheckTask(
     protected val sessionKey: SessionKey,
     private val parentEventID: EventID,
     private val messageStream: Observable<StreamContainer>,
-    private val eventBatchRouter: MessageRouter<EventBatch>
+    private val eventBatchRouter: MessageRouter<EventBatch>,
+    private val resultsStorage: ResultsStorage
```
OptimumOpium, line 87: I think it will be better if we move it into the rule configuration. We can hide ID knowledge from rule. We can pass the lambda `(EventStatus) -> Unit` instead of `ResultsStorage`

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -369,6 +379,9 @@ abstract class AbstractCheckTask(
                 .build())
         } finally {
             RuleMetric.decrementActiveRule(type())
+            if (ruleConfiguration.ruleId != 0L) {
+                resultsStorage.putResult(ruleConfiguration.ruleId, EventStatus.forNumber(refs.rootEvent.status.ordinal))
```
OptimumOpium, line 383: Why do you take the root event's status? It almost always will be PASSED. We need to check all produced events and find if any of them has FAILED status. If it is the case the rule has failed status. Otherwise, it passed

src/main/kotlin/com/exactpro/th2/check1/rule/RuleFactory.kt
```diff
@@ -101,7 +108,7 @@ class RuleFactory(
                 }
             }
 
-    fun createSequenceCheckRule(request: CheckSequenceRuleRequest, isChainIdExist: Boolean): SequenceCheckRuleTask =
+    fun createSequenceCheckRule(request: CheckSequenceRuleRequest, isChainIdExist: Boolean, ruleId: Long): SequenceCheckRuleTask =
```
OptimumOpium, line 111: I think we need to make ruleID optional in all methods

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -145,7 +151,11 @@ abstract class AbstractCheckTask(
     @Volatile
     protected var started = false
 
-    protected fun createRootEvent() = Event.from(submitTime).description(description)
+    class RuleRootEvent(timestamp: Instant) : Event(timestamp) {
+        val status: Status get() = status
+    }
```
OptimumOpium, line 156: I don't quite understand why we need this class. What was wrong with the previous implementation?

OptimumOpium (changes requested):
Please, add tests for new functionality. Maybe we should move it to the CollectorService partially. It will also allow us to hide the result storage from gRPC server implementation

src/main/kotlin/com/exactpro/th2/check1/CollectorService.kt
```diff
@@ -116,24 +129,65 @@ class CollectorService(
             null
         }
 
-        eventIdToLastCheckTask.compute(CheckTaskKey(chainID, request.connectivityId)) { _, value ->
+        rules.compute(ruleId) { _, value ->
```
OptimumOpium, line 132: Why have you changed that? We need to be able to create a chain with rules. By now we use only rule ID that is unique for every rule

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -369,6 +376,13 @@ abstract class AbstractCheckTask(
                 .build())
         } finally {
             RuleMetric.decrementActiveRule(type())
+
+            // We need to convert event to protobuf here because
+            // getAggregatedStatus() method is 'protected' in
+            // com.exactpro.th2.common.event.Event class
+            val status = refs.rootEvent.toProto(null).status
+            result.complete(status)
```
OptimumOpium, line 384: I think we can change the `publishEvent` to return aggregated event status (it produces batches and we can check them)

OptimumOpium (changes requested):

src/main/kotlin/com/exactpro/th2/check1/CollectorService.kt
lumber1000: Because CheckTaskKey keys have never been used. And this map was used as simple collection (values were accessed only through an iterator).

lumber1000: Because CheckTaskKey keys have never been used. And this map was used as simple collection (values were accessed only through an iterator).

OptimumOpium: It is used to compute the same key when a new rule with the same chain ID and connectivity ID is submitted (the new rule will be added to the existing rule to start right after the previous one is finished). Please, return the map back.

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -81,7 +85,10 @@ abstract class AbstractCheckTask(
     private val eventBatchRouter: MessageRouter<EventBatch>
 ) : AbstractSessionObserver<MessageContainer>() {
 
-    protected open class Refs(val rootEvent: Event)
+    protected open class Refs(
+        val rootEvent: Event,
+        val onTaskFinished: ((EventStatus) -> Unit)?
```
OptimumOpium, line 90: Why don't we add it to `RuleConfiguration`? I don't see any benefits from setting that parameter in every rule. Also, I think it should not be nullable. We can use an empty lambda by default

OptimumOpium (changes requested):
Also, please, add tests for new functionality

src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
lumber1000: since we have a container for references, it is appropriate to use it for all references that are not required after the rule is completed

OptimumOpium: Okay, I agree with you. Let's keep it this way

src/main/kotlin/com/exactpro/th2/check1/CollectorService.kt
```diff
@@ -90,50 +102,125 @@ class CollectorService(
...
+                .apply { addToChainOrBegin(prevTask, request.checkpoint) }
                 .run { silenceCheckTask?.also { subscribeNextTask(it) } ?: this }
         }
-        return chainID
+
+        if (resultFuture != null) {
+            ruleIdToResult[ruleId] = if (silenceFuture == null)
+                resultFuture
+            else
+                resultFuture.thenApply { if (it.first == EventStatus.SUCCESS) silenceFuture.get() else it }
```
OptimumOpium, line 163: I think you should use `thenCompose` instead of `thenApply` here

OptimumOpium (commented):
Please, add tests for functionality

src/test/kotlin/com/exactpro/th2/check1/CollectorServiceTest.kt
```diff
@@ -0,0 +1,117 @@
+package com.exactpro.th2.check1
```
OptimumOpium, line 1: Copyright is missing

src/test/kotlin/com/exactpro/th2/check1/CollectorServiceTest.kt
```diff
@@ -0,0 +1,117 @@
...
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.whenever
+import java.time.Instant
+import java.time.temporal.ChronoUnit
+import kotlin.system.measureTimeMillis
+
+class CollectorServiceTest {
+    private val eventRouterStub: MessageRouter<EventBatch> = mock {}
+    private val messageRouterStub: MessageRouter<MessageBatch> = mock<MessageRouter<MessageBatch>> {}
+        .apply { whenever(subscribeAll(any())).thenReturn(mock()) }
```
OptimumOpium, line 28: I think it is better to do this way
```suggestion
    private val messageRouterStub: MessageRouter<MessageBatch> = mock<MessageRouter<MessageBatch>> {
      on { subscribeAll(any()) }.thenReturn(mock())
    }
```

src/test/kotlin/com/exactpro/th2/check1/CollectorServiceTest.kt
```diff
@@ -0,0 +1,117 @@
...
+
+        Thread.sleep(700)
+        collector.verifyCheckRule(request) // initiate cleanup
+        val result2 = collector.getRuleResult(id, REQUEST_TIMEOUT_NANO)
+        assertEquals(CollectorService.RuleResult.NOT_FOUND, result2) // result should be deleted
+    }
+
+    companion object {
+        private const val REQUEST_TIMEOUT_NANO = 500_000_000L
+        private const val CLEANUP_OLDER_THAN_MILLIS = 500L
```
OptimumOpium, line 115: I think you can set the cleanup timeout in 100 mills and make shorter delays in your last test

OptimumOpium (changes requested):

OptimumOpium (approved):

OptimumOpium (changes requested):
Please, add changes description to the README file

src/test/kotlin/com/exactpro/th2/check1/rule/check/TestCheckRuleTask.kt
```diff
@@ -149,8 +148,7 @@ internal class TestCheckRuleTask : AbstractCheckTaskTest() {
         assertEquals(1, eventList.size)
         assertEquals(FAILED, eventList[0].status)
 
-        verify(onTaskFinishedMock).invoke(FAILED)
-        verifyNoMoreInteractions(onTaskFinishedMock)
```
OptimumOpium: Why have you removed `verifyNoMoreInteractions`?

lumber1000: I used VerificationMode.only() instead.

OptimumOpium: Okay, got it

OptimumOpium (approved):

PR #199, opened 2023-05-18
Th2 4906

Nikita-Smirnov-Exactpro:


src/main/kotlin/com/exactpro/th2/check1/MessageWrapper.kt
```diff
@@ -0,0 +1,64 @@
...
+    private val proto: Message
+) : MessageWrapper {
+    override val id: MessageID
+        get() = this.proto.metadata.id
+    override val messageType: String
+        get() = this.proto.messageType
+    override val properties: Map<String, String> = this.proto.metadata.propertiesMap
+
+    override val message: IMessage by lazy { PROTO_CONVERTER.fromProtoMessage(this.proto, false) }
+    override val metadata: IMessage by lazy { VerificationUtil.toSailfishMessage(properties) }
```
OptimumOpium, line 63: Is it really that heavy? This creates an additional object instead of a property. And that object already holds the value. Maybe we can just use the value directly

Nikita-Smirnov-Exactpro: I'm not sure ) We create several Map when create Sailfish message in another hand we create thread-safe wrapper for value when use lazy.
metadata message isn't useful for each message put into cache. I think lazy wrapper is chipper in this case

OptimumOpium: Okay, got it. I agree with you

README.md
```diff
@@ -242,6 +242,11 @@ The `th2_check1_active_tasks_number` metric separate rules with label `rule_type
 
 + Support for listening to messages using th2 transport protocol from a queue
 
+#### Changed:
+
++ Migrated to common `5.???`
++ Migrated to sailfish-utils `4.???`
```
OptimumOpium, line 248: Why don't set the current versions from `gradle.properties` in those libs? We can forget to update them here if we update the actual dependencies. And we can forget to update `???` when we know the exact versions. So, why don't set the version that is most like will be here in the end right now

src/main/kotlin/com/exactpro/th2/check1/CollectorService.kt
```diff
@@ -72,14 +77,30 @@ class CollectorService(
...
-                        SessionKey(message.metadata.id.bookName, connectionId.sessionAlias, direction)
-                    }.also(BufferMetric::processMessage)
+        protoSubscriberMonitor =
+            checkNotNull(protoMessageRouter.subscribeAll({ _: DeliveryMetadata, batch: MessageBatch ->
+                batch.messagesList.forEach {
+                    mqSubject.onNext(ProtoMessageWrapper(it))
                 }
+            })) { "Can not subscribe for listening protobuf messages" }
+        transportSubscriberMonitor =
+            checkNotNull(transportMessageRouter.subscribeAll({ _: DeliveryMetadata, batch: GroupBatch ->
```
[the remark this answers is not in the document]
Nikita-Smirnov-Exactpro: The current implementation of th2 transport parsed message doesn't support null value `body: Map<String, Any>`. Do we want to implement this feature?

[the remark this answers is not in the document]
OptimumOpium: I think we MUST do that

OptimumOpium (approved):

src/main/kotlin/com/exactpro/th2/check1/CollectorService.kt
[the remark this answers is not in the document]
Nikita-Smirnov-Exactpro: Implemented both request:

1. `This might cause an exception if no pins with transport-group are specified. So, maybe, we should make the subscription optional for transport protocol`
2. `The current implementation of th2 transport parsed message doesn't support null value body: Map<String, Any>. Do we want to implement this feature?` ... `I think we MUST do that`

OptimumOpium (approved):

build.gradle
```diff
@@ -207,10 +203,7 @@ dependencies {
 
     implementation "org.slf4j:slf4j-api"
 
-    implementation("com.exactpro.sf:sailfish-core:${sailfishVersion}") {
-        exclude group: 'com.fasterxml.jackson.dataformat', module: 'jackson-dataformat-yaml'
-        // because of the vulnerability
-        exclude group: 'junit', module: 'junit' // because of the vulnerability
+    implementation("com.exactpro.sf:sailfish-common") {
         exclude group: 'com.exactpro.mina', module: 'apache-mina-core' // because of the vulnerability
```
OptimumCode, line 207: Do we still have apache mina in the transitive dependencies?

src/main/java/com/exactpro/th2/check1/event/VerificationEntryUtils.java
```diff
@@ -92,14 +91,14 @@ private static VerificationStatus toVerificationStatus(StatusType statusType) {
...
-        default:
-            throw new IllegalArgumentException("Unsupported status type '" + statusType + '\'');
+            case NA:
+                return VerificationStatus.NA;
+            case FAILED:
+                return VerificationStatus.FAILED;
+            case PASSED:
+                return VerificationStatus.PASSED;
+            default:
+                throw new IllegalArgumentException("Unsupported status type '" + statusType + '\'');
```
OptimumCode, line 101: I believe we must stop changing formatting in the places where we don't even change the code...

OptimumCode (approved):

src/main/java/com/exactpro/th2/check1/event/VerificationEntryUtils.java
Nikita-Smirnov-Exactpro: Thank you for your comment, I have changed IDE configuration

OptimumCode (approved):

PR #203, opened 2023-10-19
[TH2-1686] Migrate to limited number of threads for rules execution

OptimumCode:
We should limit the number of threads for rules executions. This change introduces a common executor for running rules. But we can also introduce a pool of single-thread executors in order to have more than one thread for all tasks (we cannot use an executor with several threads because the rule expects to be executed on a single thread but multi-thread executor cannot guarantee that)


src/main/kotlin/com/exactpro/th2/check1/rule/AbstractCheckTask.kt
```diff
@@ -378,15 +377,15 @@ abstract class AbstractCheckTask(
             )
             status(PASSED)
             type("ruleStartPoint")
-            if (lastSequence != DEFAULT_SEQUENCE) {
-                messageID(sessionKey.toMessageID(lastSequence))
+            if (checkpointTimestamp != null && !Timestamp.getDefaultInstance().equals(checkpointTimestamp) && lastSequence != DEFAULT_SEQUENCE) {
```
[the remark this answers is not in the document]
Nikita-Smirnov-Exactpro: I thought about this way, but checkpointTimestamp is null able and I haven't search how to tell kotlin that after my isValid check pass, checkpointTimestamp isn't null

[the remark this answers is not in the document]
OptimumCode: There is a contract functionality in Kotlin. You can use it to specify a method contract (like what you described)

OptimumCode (commented):
Looks good to me

src/main/kotlin/com/exactpro/th2/check1/utils/ExecutorPool.kt
```diff
@@ -0,0 +1,59 @@
...
+        )
+    }
+
+    private val firstExecutor: ExecutorService = executors.first()
+
+    private val executorSupplier: () -> ExecutorService = when (nExecutors) {
+        1 -> { { firstExecutor } }
+        else -> { { executors.minBy { it.taskCount } } }
+    }
+    fun get(): ExecutorService = executorSupplier.invoke()
```
OptimumCode, line 52: Maybe I don't see something but I think we can simplify this
```suggestion
    fun get(): ExecutorService = when (nExecutors) {
        1 -> firstExecutor
        else -> executors.minBy { it.taskCount }
    }
```

src/main/kotlin/com/exactpro/th2/check1/utils/ExecutorPool.kt
```diff
@@ -0,0 +1,59 @@
...
+    init {
+        require(threadName.isNotBlank()) {
+            "'threadName' can't be blank, actual '$threadName'"
+        }
+        require(nExecutors > 0) {
+            "'nExecutors' must be positive, actual '$nExecutors'"
+        }
+    }
+
+    private val executors = (0 until nExecutors).map { index ->
```
OptimumCode, line 35: Just want to say that there is another option in kotlin to get an opened range - `0..<nExecutors`. It is experimental and will require opt-in annotation but it is still an option. It is up to you to keep the current one or try a new one

Nikita-Smirnov-Exactpro: Thanks

OptimumCode (commented):
Looks good to me

PR #206, opened 2024-01-16
Multisubmit rules

lumber1000:


build.gradle
```diff
@@ -196,9 +196,9 @@ signing {
 
 dependencies {
     api platform('com.exactpro.th2:bom:4.5.0')
-    implementation 'com.exactpro.th2:grpc-check1:4.3.0-dev'
-    implementation 'com.exactpro.th2:common:5.6.0-dev'
-    implementation 'com.exactpro.th2:common-utils:2.2.0-dev'
+    implementation 'com.exactpro.th2:grpc-check1:4.4.0-multisubmit-dv4-7567351808-82885f6-SNAPSHOT'
+    implementation 'com.exactpro.th2:common:5.7.1-dev'
```
OptimumCode, line 200: 5.7.2-dev is the latest version, I think we can use it here

build.gradle
```diff
@@ -279,4 +279,4 @@ licenseReport {
     ]
     excludeOwnGroup = false
     allowedLicensesFile = new URL("https://raw.githubusercontent.com/th2-net/.github/main/license-compliance/gradle-license-report/allowed-licenses.json")
-}
+}
```
OptimumCode, line 282: Do we need to change this?

src/main/java/com/exactpro/th2/check1/Check1Handler.java
```diff
@@ -109,7 +116,11 @@ public void submitCheckSequenceRule(CheckSequenceRuleRequest request, StreamObse
                 if (logger.isInfoEnabled()) {
                     logger.info("Submitting sequence rule for request '" + MessageUtils.toJson(request) + "' started");
                 }
+
+                var start = System.currentTimeMillis();
                 ChainID chainID = collectorService.verifyCheckSequenceRule(request);
+                logger.info("verifyCheckSequenceRule completed in " + (System.currentTimeMillis() - start) + "ms");
```
OptimumCode, line 122:
```suggestion
                logger.info("verifyCheckSequenceRule completed in {} ms", System.currentTimeMillis() - start);
```

src/main/java/com/exactpro/th2/check1/Check1Handler.java
```diff
@@ -174,4 +185,64 @@ public void submitNoMessageCheck(NoMessageCheckRequest request, StreamObserver<N
...
     }
+
+    @Override
+    public void submitMultipleRules(MultiRulesRequest request, StreamObserver<MultiRulesResponse> responseObserver) {
+        if (logger.isInfoEnabled()) {
+            logger.info("Rules list for request '{}' started", MessageUtils.toJson(request));
+        }
+
+        final boolean isResponseNeeded = !request.getOmitResponse();
+        final MultiRulesResponse.Builder responseBuilder = MultiRulesResponse.newBuilder();
```
OptimumCode, line 196: In all our projects we don't use `final` for local variables. Probably, we should follow the common conversion and don't start doing that here

src/main/java/com/exactpro/th2/check1/Check1Handler.java
```diff
@@ -174,4 +185,64 @@ public void submitNoMessageCheck(NoMessageCheckRequest request, StreamObserver<N
...
+
+        if (!isResponseNeeded) {
+            responseObserver.onNext(responseBuilder.build());
+            responseObserver.onCompleted();
+        }
```
OptimumCode, line 201: Honestly, this looks a bit strange... I would think that having two methods (one returns MultiRulesResponse and another one returns Empty in response) would be better in terms of clarity for users

src/main/kotlin/com/exactpro/th2/check1/configuration/Check1Configuration.kt
```diff
@@ -27,12 +30,12 @@ class Check1Configuration(
...
     val cleanupOlderThan: Long = 60L,
 
-    @field:JsonProperty(value = "max-event-batch-content-size", defaultValue = "1048576")
-    val maxEventBatchContentSize: Int = 1048576,
-
     @field:JsonProperty(value = "cleanup-time-unit", defaultValue = "SECONDS")
     val cleanupTimeUnit: ChronoUnit = ChronoUnit.SECONDS,
 
+    @field:JsonProperty(value = "cleanup-interval-millis", defaultValue = "1000")
+    val cleanupIntervalMillis: Long = 1000L,
```
OptimumCode, line 37: Maybe we could compute this based on the submitted tasks. We could find the earliest submitted task, take its submission time and add `cleanupTime` to that time. This will be the next time we should check for tasks to clean up.
The introduction of another parameter that should be configured by the user does not bring any benefits to us.

build.gradle
lumber1000: some unit tests fail with it

OptimumCode: Okay. Did you check which unit tests and why they failed?

lumber1000: fixed (missed `scope` value in tests)

src/main/java/com/exactpro/th2/check1/Check1Handler.java
lumber1000: added `postMultipleRules` method

src/main/kotlin/com/exactpro/th2/check1/configuration/Check1Configuration.kt
lumber1000: This will not reduce number of cleanups significantly. We submit thousands rules a second and always have some rules for cleanup. We could use fixed interval of 1 second and remove config parameter.

OptimumCode: I am sorry, but I don't see any correlation between the 1k rule/s and the cleanup timeout. If all these 1k rules are submitted with unique chain IDs we cannot clean up them before the cleanup time has elapsed. Otherwise, users won't be able to use them for chaining other rules

OptimumCode (approved):

src/main/kotlin/com/exactpro/th2/check1/configuration/Check1Configuration.kt
lumber1000: If we'll submit 1k rule/s we will get about 1k/s timeout expirations. So, average time to expiration of earliest submitted task after previous cleanup will be ~1ms. And I'm trying to make interval between cleanupds at least 1 sec (or `cleanup-interval-millis`).

OptimumCode: Okay, I believe I got you. So the concern here is that if rules are submitted with a rate of 1k rule/s eventually, all these rules will be cleaned up one by one and you want to introduce a minimal delay between two clean ups. Please, correct me here.
I think we need to rename the parameter to 'cleanup-cooldown-ms' or 'cleanup-quiet-period-ms' because 'cleanup-interval' looks more like a cleanup scheduling parameter

OptimumCode (approved):

PR #209, opened 2024-10-30
[th2-5248] Passed noMessageCheckExecutionStop event in noMessageRule check when message_timeout = 0 and timeout expired 

Nikita-Smirnov-Exactpro:


src/main/kotlin/com/exactpro/th2/check1/rule/nomessage/NoMessageCheckTask.kt
```diff
@@ -48,6 +48,8 @@ class NoMessageCheckTask(
     onTaskFinished: ((EventStatus) -> Unit) = EMPTY_STATUS_CONSUMER
 ) : AbstractCheckTask(ruleConfiguration, startTime, sessionKey, parentEventID, messageStream, eventBatchRouter) {
 
+    private val hasMessageTimeout = ruleConfiguration.taskTimeout.messageTimeout != 0L
```
OptimumCode, line 51: Just to avoid having an additional field in the class
```suggestion
    private val hasMessageTimeout: Boolean
      get() = ruleConfiguration.taskTimeout.messageTimeout != 0L
```

OptimumCode (approved):

OptimumCode (approved):

